PR #1, opened 2017-07-17
fixed wrong module path in pynetbox/lib/__init__.py

h-otter:
There was wrong module path in `pynetbox/lib/__init__.py`, so pynetbox cannot work well on python 3.


zachmoody:
😖  ah, missed this one when I was getting rid of relative imports other places, thanks @h-otter! Will have a new version pushed shortly.

PR #2, opened 2017-08-03
Correct doc string example

lamiskin:
I believe nb.dcim.device.all(), needs to be plural devices.


zachmoody:
Yep. Thanks for the fix!

Issue #22, opened 2017-12-14
pynetbox should support patching


explody:
Using pynetbox, we quickly ran into scenarios where we needed to bulk-update a resource from a python dict or similar multi-value source.  But, we noticed that there didn't seem to be PATCH support, nor a way to update a resource, aside from updating an object's attributes one by one, then calling save().

It would be extremely useful to be able to do two things:

1) Update a resource, given an ID and update data, without previously fetching it as an object, i.e.
```
netbox.dcim.devices.update(1, {'key': 'value', 'otherkey': 'othervalue'})
```

2) Update a retrieved object with a dict
```
dev = netbox.dcim.devices.get(1)
dev.update({'key': 'value', 'otherkey': 'othervalue'})
dev.save()
```

cweir-tacc:
I have now running into a similar issue and would like an elegant solution.

zachmoody:
So the latter approach will be in the 4.0.0 release. Opted not to do a method off Endpoint because you likely wouldn't know the `id` beforehand, and something like:
```
for i in netbox.dcim.devices.filter(blah='blah'):
    i.update({{'key': 'value', 'otherkey': 'othervalue'})
```
Seems more natural and would lead to fewer queries. If I'm missing something though feel free to open a new issue and we can discuss it.

Issue #21, opened 2017-12-14
pynetbox ssl validation should be configurable


explody:
As TLS certs, particularly in test pipelines, may be generated from private CAs or self-signed, pynetbox should allow for _not_ verifying SSL certs, if explicitly set to do so.  The current behavior uses Requests' default, which is to require a valid cert and fail if it cannot be verified.

Ichabond:
Requests allow you to export `REQUESTS_CA_BUNDLE` (http://docs.python-requests.org/en/master/user/advanced/#ssl-cert-verification). This should fix the dev issue, without giving users a shotgun when it comes to security.

explody:
Defaulting an available configuration to the preferred setting while still making it available for the user to override the behavior does not seem like much of a shotgun, IMHO.

Also, I don't believe this is equivalent function.  With access to build a bundle of CA certs, one could add them to the system bundle without requiring any specific config in the api client.

The use case we addressed is more of:  netbox instance is auto created with a self-signed cert as part of a pipeline while a client on a separate VM needs to access this transient instance over SSL.  The client never has direct access to the cert from which to build a bundle.  We could also just as easily remove SSL entirely or re-use a predictable cert, but neither is better practice IMHO. 

Besides, not verifying a server cert is really only a security issue when one does not control both ends of the connection and one has to trust _someone else_ ;)

zachmoody:
Totally agree that every effort should be made to make your SSL infrastructure sane, but enforcing that  in an obscure python lib like this is seems like a bit of an overreach, imo.   :smile:

Dropped some comments in your PR, but overall I don't think it's a big deal to let the user toggle cert verification.

PR #23, opened 2017-12-14
Patch and ssl. Addresses #21 and part of #22

explody:
This was originally going to be two separate PRs, but the update feature also required the ssl change so we ended up mashing them together.

### SSL for #21 
This part is pretty simple.  It just adds an additional argument to several methods and the constructors of classes that end up making HTTP calls, to support Requests' 'verify' argument, implemented here as 'ssl_verify' to be more explicit about what the option does.  It defaults to 'True', which is the current default behavior, or will accept 'False' or a string that is a path to a CA cert (see requests docs).  Usage is simple:

```
netbox = pynetbox.api(
    myurl,
    token=mytoken,
    ssl_verify=False
)
```

Reference is here: http://docs.python-requests.org/en/master/user/advanced/#ssl-cert-verification

### PATCH for part 1 of #22 
This portion of the PR adds a patch() method to Request.  It's a near copy of put(), merely using the different HTTP method.

It also adds an update() method to Endpoint, which implements the first suggestion in #22, a way to update a resource given a dict, without retrieving it as an object first.  

I did _not_ implement an update() method in the Record class.  Though it's very doable, there were some design considerations that probably need to be made - i.e. how to handle nested Records, etc.  It won't be hard, but I figured it would warrant a conversation first.

Let me know if this looks useful, and if you'd want any changes to what's here.

Cheers!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,7 +14,6 @@ dist/
 downloads/
 eggs/
 .eggs/
-lib/
 lib64/
 parts/
 sdist/
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@ downloads/
 eggs/
 .eggs/
 lib64/
+lib/
 parts/
 sdist/
 var/
--- a/pynetbox/api.py
+++ b/pynetbox/api.py
@@ -80,7 +80,7 @@ class Api(object):
     """
 
     def __init__(self, url, token=None, private_key=None,
-                 private_key_file=None, version=None):
+                 private_key_file=None, version=None, ssl_verify=True):
         if private_key and private_key_file:
             raise ValueError(
                 '"private_key" and "private_key_file" cannot be used together.'
@@ -92,6 +92,7 @@ class Api(object):
             "private_key": private_key,
             "private_key_file": private_key_file,
             "base_url": base_url,
+            "ssl_verify": ssl_verify,
         }
 
         if self.api_kwargs.get('private_key_file'):
@@ -99,13 +100,14 @@ class Api(object):
                 private_key = kf.read()
                 self.api_kwargs.update(private_key=private_key)
         if not version:
-            version = Request(base=base_url).get_version()
+            version = Request(base=base_url, ssl_verify=ssl_verify).get_version()
         self.api_kwargs.update(version=version)
 
         req = Request(
             base=base_url,
             token=token,
-            private_key=private_key
+            private_key=private_key,
+            ssl_verify=ssl_verify
         )
         if token and private_key:
             self.api_kwargs.update(session_key=req.get_session_key())
--- a/pynetbox/api.py
+++ b/pynetbox/api.py
@@ -100,7 +100,8 @@ class Api(object):
                 private_key = kf.read()
                 self.api_kwargs.update(private_key=private_key)
         if not version:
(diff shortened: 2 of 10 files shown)


zachmoody:
Nice! Thanks for the PR with tests. PATCH support by some means has been a thing I've wanted to add for a while now. Didn't know that SSL validation was a problem with self-signed certs. Thanks for catching that.  

So you can probably tell that pynetbox's API is _heavily_ influenced by the Django ORM. The thought I'd had for PATCH support off Endpoints was similar to the `update()` method on QuerySets. Mainly so we can take advantage of updating collections of objects returned by filter() and all(). E.g. `nb.dcim.devices.filter(site='somewhere', role='somerole').update(status=1)` to change the status field on all devices in a certain role at a site. Of course, that'd require `all()` and `filter()` to return an object that's very roughly similar to a QuerySet where we can implement `.update()` (and probably `.delete()` too). 

I'm just not sure that hanging an `.update()` off `Endpoints` that only takes an ID is the best experience. I usually don't know the PK of the record I want to update without making a GET first, and at that point, calling `update()` off `Record` would seem a bit more logical. If I'm missing something here, let me know. I'm kind of thinking about it from the other direction. If we had `.update()` on `Record` objects we could simply make the same method off `Endpoints` or whatever intermediate object we end up creating just call that.

Happy to merge the SSL verify stuff though, but maybe with a test to make sure we're actually passing the arg to requests. Api args get passed around quite a bit and I'd like to make sure this was actually getting there since it's not something I caught before now.

explody:
I totally agree w/ you, actually.  I perceived two experiences for doing updates (just using `device` as an example):

```python
# #1
netbox.dcim.devices.update(some_args_including_device_identifier)
```

```python
# #2 
device = netbox.dcim.devices.get(1)
device.update(device_data)
```

With the presumption that `device.update()` and `netbox.dcim.devices.update()` are essentially calling the same `patch()` code.  I think approach 2 is way more desirable but also had more nuance to how it could be implemented (exactly to your point about QuerySets, etc) so I did approach 1 first, since it's just a paper-thin wrapper around `patch()`.   The `Endpoint.update()` method could be removed but it did seem appropriate alongside get/create/filter/all, just not a complete implementation w/o `Record.update()` as well.

With that in mind, do you think this would be useful as a starting point or would you prefer the ssl_verify feature independently?

zachmoody:
Yeah, let's get the cert verification toggle merged sooner as a separate PR. If you can include a test for making sure the requests lib gets passed the right options too, that'd be a bonus.

As far as PATCH support, I think everything could revolve around an `update()` method on `Record`. You could immediately use it like you show in the second example, but we could add niceties like either example 1 or like QuerySet's later that just called `.update()` on the Records themselves. All with backwards compatible API changes too.

The only thing that might stymie that approach is how NetBox's bulk creation feature would work with PATCH. I wouldn't want to make _n_ requests when we could just pass a list of patches in one request. Either way, that's a bit of a tangent, we'd need patch support on Records before any of that.

I haven't gotten too far into the details of that, especially as you mentioned, wrt nested objects. I think we could use roughly the same logic as is used for PUT and POST, but would have to get into it to make sure. If you feel like tackling it, by all means, have at it. It might be a few more weeks before I can make a pass at it.

PR #25, opened 2017-12-18
Add support for _choices

### LINKED ISSUE #24: Add support for the dcim/_choices endpoint
Netbox has a `_choices` endpoint, which is currently not supported, as it gets remapped to `-choices` in `endpoint.py` (https://github.com/digitalocean/pynetbox/blob/master/pynetbox/lib/endpoint.py#L62), and it doesn't map to the default `Record` structure.

Update the application so it no longer remaps endpoints beginning with `_`, and add a specific class `Choice` that only has an `items()` function to retrieve the results.

Ichabond:
This adds initial support for the `dcim/_choices` endpoint. This fixes #24 

I added a seperate `Choice` class, to deal with the unique return values. Added an exception for the `_choices` endpoint to map to this class.

### DIFF
--- a/pynetbox/lib/__init__.py
+++ b/pynetbox/lib/__init__.py
@@ -1,3 +1,3 @@
 from pynetbox.lib.endpoint import Endpoint
-from pynetbox.lib.response import Record, IPRecord
+from pynetbox.lib.response import Record, IPRecord, Choice
 from pynetbox.lib.query import Request, RequestError
--- a/pynetbox/lib/endpoint.py
+++ b/pynetbox/lib/endpoint.py
@@ -16,7 +16,7 @@ limitations under the License.
 from collections import defaultdict
 
 from pynetbox.lib.query import Request
-from pynetbox.lib.response import Record, IPRecord
+from pynetbox.lib.response import Record, IPRecord, Choice
 
 CACHE = defaultdict(list)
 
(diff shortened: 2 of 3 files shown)


zachmoody:
:wave: Thanks for the PR. I'd been playing around with the idea of using the data from `_choices` to make places we have to use the values more human-friendly (e.g. mostly kwargs for the various Record and Endpoint methods), but not necessarily exposing the endpoint itself. Just curious if that would address what you were trying to do or there's some uses I'm missing. 

I didn't have a lot of time to play around with your branch, but it looks like all I could get are the keys from the NetBox response. Was that all you were looking to do? Some tests (need these anyways) and more docs would definitely help me better understand the goal.

Ichabond:
Hey,

The values of _choices are needed when creating or modifying an interface to set the interface form-factor. I think it would still be a good idea to expose what the options are, instead of just internally mapping a human-readable label to the value, although that is a good idea as well.

If you do the following:
`api.dcim._choices.get('interface:form_factor')`, you get the label and values, which is currently all that's really needed. Maybe use the label as a dictionary key, and merge the list of dictionaries into 1 dictionary instead?

zachmoody:
Got some more time to play with it. I was using `all()` and getting some odd behavior. It returned a list of `Choice` objects, but `.items()` just returned the key as a string, but `.get()` works though. I'm kinda thinking an `Endpoint` object might not be the right container for this stuff. Calling any of the other method ends badly. 

Had a half-baked idea. Maybe a `.choices()` Endpoint method that'd by default return the choices for the endpoint it's called from (maybe pass a kwarg to other endpoint's choices, if needed)? Could maybe use that to provide interpolation for values passed to `create()`, and pass it down to the `Record` objects for the same thing. Will give it a little more thought though.

At the very least, I'm thinking we should subclass Endpoint and raise `NotImplemented` where it makes sense.

vzaletny:
Hi guys.

How about support for _choices in pynetbox?

Issue #32, opened 2018-02-07
full_details doesn't add token


funzoneq:
I'm using pynetbox version 3.0.1 and netbox with LOGIN_REQUIRED = True and I see the following behaviour:

```
import requests
import logging
from pprint import pprint

# These two lines enable debugging at httplib level (requests->urllib3->http.client)
# You will see the REQUEST, including HEADERS and DATA, and RESPONSE with HEADERS but without DATA.
# The only thing missing will be the response.body which is not logged.
try:
    import http.client as http_client
except ImportError:
    # Python 2
    import httplib as http_client
http_client.HTTPConnection.debuglevel = 1

# You must initialize logging, otherwise you'll not see debug output.
logging.basicConfig()
logging.getLogger().setLevel(logging.DEBUG)
requests_log = logging.getLogger("requests.packages.urllib3")
requests_log.setLevel(logging.DEBUG)
requests_log.propagate = True

url = 'https://netbox.example.com'
token = 'abcdefghijklmnopqrstuvwxyz'

try:
    import pynetbox
    from netaddr import *
except ImportError, e:
    sys.exit("Please install required python modules: pynetbox netaddr")

try:
    netbox = pynetbox.api(url, token=token)
except AttributeError:
    sys.exit("Failed to connect to netbox instance")

for host in netbox.virtualization.virtual_machines.filter(status=1):
    pprint(host)
    pprint(vars(host))
```

```
INFO:requests.packages.urllib3.connectionpool:Starting new HTTPS connection (1): netbox.example.net
send: 'GET /api/ HTTP/1.1\r\nHost: netbox.example.net\r\nConnection: keep-alive\r\nAccept-Encoding: gzip, deflate\r\nAccept: */*\r\nUser-Agent: python-requests/2.10.0\r\n\r\n'
reply: 'HTTP/1.1 200 OK\r\n'
header: Server: nginx/1.10.3 (Ubuntu)
header: Date: Wed, 07 Feb 2018 09:38:23 GMT
header: Content-Type: application/json
header: Content-Length: 353
header: Connection: keep-alive
header: Allow: GET, HEAD, OPTIONS
header: X-Frame-Options: SAMEORIGIN
header: API-Version: 2.2
header: Vary: Accept, Cookie
header: P3P: CP="ALL DSP COR PSAa PSDa OUR NOR ONL UNI COM NAV"
header: X-Cache-Status: MISS
[...]

PR #33, opened 2018-02-07
full_details doesn't add token #32

funzoneq:
Fix for bug #32


zachmoody:
Yep, looks good. Thanks for the PR.

PR #37, opened 2018-02-27
Add string representation for new virtual chassis endpoint

lamiskin:
Netbox 2.3.0 is out with virtual chassis support.


zachmoody:
Hey, thanks @lamiskin - Haven't had time to play with the virtual-chassis stuff myself yet, but the PR looks good. Should be able to merge this and cut a new version today.

zachmoody:
Sorry for the delay, I noticed (I think you did too in some of the first commits) the `GenericTest` object is ill suited to handle an endpoint with a name like this. Tests pass, but it's querying and expecting the wrong url because the name attribute is a bit overloaded. Gonna create an issue to get that fixed up, but I don't see a reason to hold this up any longer because of that. Thanks again.

PR #53, opened 2018-04-26
fix some urlencoding problems

### LINKED ISSUE #52: Some urls are not fully urlencoded
We had some trouble with devices that have special chars in their name, such as `#`. Here's an example:

```
>>> i =  netbox.dcim.interfaces.get(name='ipmi', device='New TwinU db #2 / D')
>>> i is None
True
```

The access log looks like this 
```
"GET /api/dcim/interfaces/?name=ipmi&device=New%20TwinU%20db%20 HTTP/1.1"
```

It's clear the rest of the device name is getting cut off because the `#` was not encoded.

mchrobak:
fixes #52 
- This fixes some problems with urls that were not encoded, causing devices with characters in the name (like `#`) to not be retrieved through pynetbox.

### DIFF
--- a/pynetbox/lib/query.py
+++ b/pynetbox/lib/query.py
@@ -25,10 +25,7 @@ def url_param_builder(param_dict):
     Creates URL paramters (e.g. '.../?xyz=r21&abc=123') from a dict
     passed in param_dict
     '''
-    param_list = []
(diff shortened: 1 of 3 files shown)


zachmoody:
:+1: Great stuff, thanks for the PR.

Issue #56, opened 2018-05-15
pip errors on new version of netaddr


2bithacker:
When doing a `pip install pynetbox`, I've been getting the following recently:

```
pynetbox 3.3.1 has requirement netaddr==0.7.18, but you'll have netaddr 0.7.19 which is incompatible.
```

Forcing the netaddr version by doing `pip install netaddr==0.7.18` works fine.

Probably just need a version bump for requirements.

PR #57, opened 2018-05-16
`netaddr` usage

mandarg:
This is more of an issue than a pull request...We noticed that Pynetbox was using `netaddr` instead of `ipaddress`, and were wondering if there's a specific reason for that. Is it more extensible, or does it provide better Py3 compatibility or something similar?

As a proof-of-concept, here's a patch that converts netbox to use `ipaddress` (built in from Python3.3+, a PyPI module for earlier version) if you want to try it out.

### DIFF
--- a/pynetbox/lib/response.py
+++ b/pynetbox/lib/response.py
@@ -13,8 +13,8 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 '''
-import netaddr
 import six
+import ipaddress
 
 from pynetbox.lib.query import Request
 
@@ -200,8 +200,8 @@ class Record(object):
             if isinstance(current_val, Record):
                 current_val = getattr(current_val, 'serialize')(nested=True)
 
-            if isinstance(current_val, netaddr.ip.IPNetwork):
-                current_val = str(current_val)
+            if isinstance(current_val, ipaddress._IPAddressBase):
+                current_val = current_val.exploded
 
             ret.update({i: current_val})
         return ret
@@ -265,7 +265,7 @@ class IPRecord(Record):
     """IP-specific Record for IPAM responses.
 
     Extends ``Record`` objects to handle replacing ip4/6 strings with
-    instances of ``netaddr.IPNetworks`` instead.
+    instances of ``ipaddress.IPv*Networks`` instead.
     """
 
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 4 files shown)


mandarg:
Just noticed that this will probably be relevant for Issue #56 as well.

zachmoody:
Hey, thanks for the PR. No reason in particular I chose netaddr over ipaddress. Just went with what I'd been using for years. I remember trying ipaddr (I think that's the py2 predecessor to ipaddress) way back when I started using python, but don't remember why I dropped it for netaddr. I want to say it had more features back then, or maybe just had a feature I needed at the time, can't remember...it was a while ago :smile:.

I'm a little hesitant to rip out netaddr mostly for what it'd mean for downstream code. It'd probably break things and would warrant a major version bump. In hindsight, I probably should've just returned them as strings so people can use whatever they like.

Happy to leave the PR open for more discussion. Those are just my opinions.

mandarg:
>  It'd probably break things and would warrant a major version bump.

Yeah, that's a fair concern. One thought is that this could go in at a point where there was a planned major version bump for other reasons anyway.

> I'm a little hesitant to rip out netaddr mostly for what it'd mean for downstream code.

Are there specific downstream projects you're worried about? One thing I noticed was that we overlooked [pynetbox](https://github.com/digitalocean/pynetbox) which seems to use `netaddr` as well. What are some other ones I should know about?

zachmoody:
> Yeah, that's a fair concern. One thought is that this could go in at a point where there was a planned major version bump for other reasons anyway.

I'm anticipating the next major version is going to land when I can have `.create()` simply "refresh" the existing object after it's called instead of returning a dict with NetBox's response. It's been blocked by a slight inconsistency in NetBox's API that I think may be resolved now. Just need some time to see that it is, and make pynetbox to behave that way.

>  Are there specific downstream projects you're worried about? If you could specify some, we could look into patching them as well.

Nothing I know of that's public, but could be a lot of random scripts that'll break.

How do you feel about just being library agnostic and returning strings? The more I think about it, the more in favor of it I am.

jord-bh:
> How do you feel about just being library agnostic and returning strings? The more I think about it, the more in favor of it I am.

I second this.

mandarg:
> How do you feel about just being library agnostic and returning strings? The more I think about it, the more in favor of it I am.

This sounds good from our usage perspective, since we can always wrap anything returned using `ipaddress`.

@zachmoody Feel free to change the issue, or drop it and create a new "Return IP addresses/ranges as strings". I can also do this if there aren't any more extensive implementation details you want to put in it.

zachmoody:
Yeah, it is a bit more extensive than just returning strings where we used to return netaddr.IPNetwork objects. The idea is to have `.create()` return relevant Record object(s) instead of just a dict response from NetBox. And those objects, if they have IPs or prefix attrs, would then just be strings. I think this would be possible in NetBox 2.4 because of the work done to unify the serializers, but I have to spend some time with it and make sure I'm not misunderstanding something.

PR #69, opened 2018-07-04
Remove setuptools-scm requirement

mandarg:
7c8e2efab8ec593b0b3d318446439f51768f5c86 removed the need for setuptools-scm, so it would be good to remove it from the `install_requires`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -17,7 +17,6 @@ setup(
         'netaddr==0.*',
         'requests==2.*',
         'six==1.*',
-        'setuptools-scm==2.*'
     ],
     zip_safe=False,
     keywords=['netbox'],


mandarg:
Actually, I guess `use_scm_version` is still [here](https://github.com/digitalocean/pynetbox/blob/master/setup.py#L11) – which would imply that `setuptools-scm` is still being used. However, in _that_ case, [their own instructions](https://github.com/pypa/setuptools_scm#setuppy-usage) seem to say that `setup_requires=['setuptools_scm']` should be the way to go.

zachmoody:
I think it's ok as long as it's just used in setup.py.

PR #74, opened 2018-07-15
Return id when name and label aren't present. 

funzoneq:
For example on api/circuits/circuits endpoint only returns id, cid. No name or label, so you would get an array that looked like:

```
[, , , , , ]
```

### DIFF
--- a/pynetbox/lib/response.py
+++ b/pynetbox/lib/response.py
@@ -100,9 +100,10 @@ class Record(object):
         return item
 
     def __str__(self):
-        return (
+        return str(
             getattr(self, 'name', None) or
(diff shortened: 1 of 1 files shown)


zachmoody:
Hey, @funzoneq, thanks for the PR. The way pynetbox typically handles this is by making a unique object that sub-classes `Record` and has its own str representation. e.g. https://github.com/digitalocean/pynetbox/blob/master/pynetbox/dcim.py#L71-L72.

It's pretty straight forward. Just need to create a `circuit.py` with the custom object and import and reference it in `api.py`.

zachmoody:
This should be fixed in `v3.4.4`

PR #81, opened 2018-07-24
Adding GitHub templates for issues and pull requests

bdlamprecht:
Using modified NetBox repository templates


bdlamprecht:
I simply copies the issue and pull request templates over from `NetBox` and modified them appropriately as that is what I was doing consistently.

Thought it would be nice to not have to keep doing that manually and would benefit others as well.

PR #96, opened 2018-10-02
Added special_endpoint to api, which allows to access strange properties

ChloeTigre:
Hello,

I just put this patch together to cover my specific need of filling our Netbox rack inventory with pynetbox.

See patch for example usage and the obvious rationale.

### DIFF
--- a/pynetbox/api.py
+++ b/pynetbox/api.py
@@ -37,6 +37,27 @@ class App(object):
             app=self.app
         )
 
+    def special_endpoint(self, **kwargs):
+        """Return special endpoint
+
+        Some resources like /dcim/racks/{id}/units/ cannot be accessed
+        otherwise
+
+        :returns: Endpoint
+        :example:
+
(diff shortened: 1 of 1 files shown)


zachmoody:
:wave: Thanks for the PR, there's already a construct for dealing with detail routes like this, that I think _should_ work here. It'd just need to be added to a custom `Rack` object in dcim.py. Here's a good example of what I'm talking about: https://github.com/digitalocean/pynetbox/blob/master/pynetbox/dcim.py#L47-L66

ChloeTigre:
Hi,

I'll give it a shot and re-submit a PR, thanks :)

ChloeTigre:
I don't see how it could be used. I tried, so to make a comparison, to find a "napalm" doc in the swagger https://netbox/api/docs/  with no success, so I don't have an opportunity to compare.

Would it be possible for another developer to take a look at this and provide guidance? I realize my way of doing stuff is not as clean as I'd expect so another opinion and example might help.

PR #100, opened 2018-10-15
Conceal the SSL-Certificate-Warning

mabazyar:
Suppresses the SSL-Certificate-Warning if the flag ssl_verify is already set to False

### DIFF
--- a/pynetbox/lib/query.py
+++ b/pynetbox/lib/query.py
@@ -17,7 +17,8 @@ import json
 from six.moves.urllib.parse import urlencode
 
 import requests
-
+from requests.packages.urllib3.exceptions import InsecureRequestWarning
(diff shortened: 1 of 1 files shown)


zachmoody:
Looks like this disables it globally, regardless of what the user passes in with `ssl_verify`. Generally speaking though, I'd rather not get in the business of suppressing warnings from dependencies. I haven't played around with invalid certs myself, but can't a client just do this in their code if they desire?

mabazyar:
Could it be conditional that general, like so:

```
if not ssl_verify: requests.packages.urllib3.disable_warnings(InsecureRequestWarning)
```

So that, for those who do not care about the cert checking, the warning is gone away...

PR #101, opened 2018-10-17
Add JSON_FIELD local_context_data.

funzoneq:
The local_context_data field isn't correctly exposed. This PR fixes that.

Netbox version v2.4.7
pynetbox version v3.4.7

### DIFF
--- a/pynetbox/lib/response.py
+++ b/pynetbox/lib/response.py
@@ -20,7 +20,7 @@ from pynetbox.lib.query import Request
 
 # List of fields that contain a dict but are not to be converted into
 # Record objects.
-JSON_FIELDS = ('custom_fields', 'data', 'config_context')
+JSON_FIELDS = ('custom_fields', 'data', 'config_context', 'local_context_data')
 
 
 def get_return(lookup, return_fields=None):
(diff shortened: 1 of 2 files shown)


pynetbox/lib/response.py
```diff
@@ -161,7 +161,7 @@ def _parse_values(self, values):
                         v = self.default_ret(v, api_kwargs=self.api_kwargs)
                 self._add_cache((k, v))
             else:
-                self._add_cache((k, v.copy()))
+                self._add_cache((k, v))
```
zachmoody: The rest looks good, but what's the reason for dropping copy here?

a31amit:
@zachmoody . Can you please do your magic here and get it to release. This will be helpful to use.

pynetbox/lib/response.py
B1ad3s: It appears that the reason for the removal of the copy is that local_context_data can be returned as types None or str which do not have the copy method. Leaving the copy when trying to return any empty local_context_data fields then breaks.

zachmoody:
Implemented in PR #197

Issue #122, opened 2018-12-06
dcim.choices() throws 403


abringenberg:
The new method choices() of the app object throws `pynetbox.core.query.RequestError: The request failed with code 403 Forbidden: {'detail': 'Authentication credentials were not provided.'}`.

Adding token and private_key from self.api to Request of api.py:48 solves the problem.

PR #123, opened 2018-12-06
add missing authentication credentials

abringenberg:
Solves #122


zachmoody:
:man_facepalming: Good catch, thanks for the PR!

PR #125, opened 2018-12-07
TypeError: sequence item 0: expected str instance.

VictorPavlushin:
If an array of integers, then we get an error:
File "/opt/ansible-py3.6/lib/python3.6/site-packages/pynetbox/core/response.py", line 329, in fmt_dict
    return k, "".join(v)
TypeError: sequence item 0: expected str instance, int found

Just convert to string.
I got an error when there are vlans on the interface.


zachmoody:
:+1: Thanks for fixing that.

PR #132, opened 2018-12-14
Сertain sequence vlan id

VictorPavlushin:
If you transfer for example 2 vlans with id 45 and 46, and then change to 4 and 546 then diff will not dispel the change.
The result in both cases will be 4546.


zachmoody:
Yeah, good point. Thanks for the PR.

PR #139, opened 2019-01-03
Single TCP Session for HTTP Requests

hSaria:
Instead of establishing an entirely new TCP session and closing it for each individual request, only one session is created. 

This approach saves a lot of time. To be more accurate, for every individual request after the first one that created the socket, the following packets will not be sent (S) or received (R):
- (S) SYN 
- (R) SYN-ACK
- (S) ACK
- (S) FIN
- (R) FIN

I've ran a test (pull list 100 times back-to-back) multiple times in a controlled environment (set the delay to 20ms in VMware) and the results are:
- One TCP session per request: 7.536s
- One TCP session for all requests: 4.712s

The savings here aren't necessarily on the bandwidth aspect, but the delay one; all of the above packets are blocking (the client/server must wait until the corresponding packet is sent/received).

Furthermore, the delay in the test was set 20ms. The higher the delay and/or the volume of requests, the worse this effect becomes.

### DIFF
--- a/pynetbox/api.py
+++ b/pynetbox/api.py
@@ -46,6 +46,7 @@ class App(object):
             return self._choices
 
         self._choices = Request(
+            api=self.api,
             base="{}/{}/_choices/".format(self.api.base_url, self.name),
             token=self.api.token,
             private_key=self.api.private_key,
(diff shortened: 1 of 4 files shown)


hSaria:
I've just realized that there already a pull-request for this... Feel free to close this one :)

PR #148, opened 2019-01-24
add requests http session to Api, to Request and usage of Request

abringenberg:
While using pynetbox for big amount of small changes in our netbox database I noticed an annoying latency. A colleague had a look to the network traffic and told me, that for every request a single http session gets initialized, used and closed.
This PR uses the onboard http sessions from requests. initialized in Api() and used by Request() and DetailEndpoint. It speeds up the communication by (re-)using a single http session per Api objects including keep-alive. The improvement is around 25%

### DIFF
--- a/pynetbox/api.py
+++ b/pynetbox/api.py
@@ -13,6 +13,8 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 """
+import requests
+
 from pynetbox.core.endpoint import Endpoint
 from pynetbox.core.query import Request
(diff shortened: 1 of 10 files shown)


dgarros:
Would be great to get this PR merged, sounds like a great addition

zachmoody:
I'd appreciate some more folks testing it on large and varied workloads. Each time I've tried (both against the django dev server and gunicorn) it's ended in, what seem to be, random tracebacks a few seconds in.

zachmoody:
Good to hear, I'll test again when I get a chance.

zachmoody:
Had a chance to test again and it went well this time. Not sure what was going on last time, but probably PEBKAC on my part. Thanks for the PR.

PR #155, opened 2019-02-28
Update endpoint.py

psiisyapk:
Convert Endpoint method "all" to generator. When you have many tenants - nb.tenancy.tenants.all() works very slow (for more than a second and more), while the generator is 3 times faster


psiisyapk:
unfortunately, my commit will never pass the tests because the function "all" now returns a generator instead of a list

zachmoody:
We may do this in the future to give more control over how many calls we make in functions like `.all()` and `.filter()` but would need a little more than just making them a generator. Closing for now.

PR #159, opened 2019-04-03
Added the ability to send additional http headers

maximebrun:
We use OIDC to authenticate and we have to replace Authentication header (Token vs Bearer).

Added the ability to send additional http headers to Netbox is a way more generic and acceptable.


markkuleinio:
I believe this PR can be closed as 5.0.x brought custom sessions. (@zachmoody, @maximebrun)

zachmoody:
Aye, closing.

Issue #161, opened 2019-04-12
Record url is not set + Endpoint is not updated


louis-oui:
Environment
Python version: 3.6.6
NetBox version: 2.5.10
PynetBox version: 4.0.6

**Steps to reproduce :**
Python 3.6.6 (default, Aug 13 2018, 18:24:23) 
[GCC 4.8.5 20150623 (Red Hat 4.8.5-28)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pynetbox
>>> api = pynetbox.api(
...     'http://YYYYYYYY',
...     token='XXXXXXXXX',
... )
>>> 
>>> db=api.dcim.device_bays.get(1)
>>> db.url
>>> db.endpoint.url
'http://YYYYYYYY/api/dcim/device-bays'
>>> 
>>> device=db.installed_device
>>> device.url
'http://YYYYYYYY/api/dcim/devices/2/'
>>> device.endpoint.url
'http://YYYYYYYY/api/dcim/device-bays'
>>> api.dcim.device_bays.get(2).delete()
True
>>> device.delete()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.6/site-packages/pynetbox/core/response.py", line 423, in delete
    return True if req.delete() else False
  File "/usr/local/lib/python3.6/site-packages/pynetbox/core/query.py", line 318, in delete
    raise RequestError(req)
pynetbox.core.query.RequestError: The requested url: http://YYYYYYYY/api/dcim/device-bays/2/ could not be found.

**Expected Behavior**
- db.url returns 'http://YYYYYYYY/api/dcim/device-bays/1'
- device.endpoint.url returns 'http://YYYYYYYY/api/dcim/devices/'
- device.endpoint.delete() do a DELETE on http://YYYYYYYY/api/dcim/devices/2/

**Observed Behavior**
- db.url is None
- device.endpoint.url 'http://YYYYYYYY/api/dcim/device-bays' (i.e. db.endpoint.url)
- seems device.delete() tries to do a DELETE using device.endpoint.url that is wrong

louis-oui:
It seems the bug is there in response.py def _parse_values(self, values) -> def list_parser(list_item) . self.endpoint is used there. In my opinion, a new endpoint should be used:

v = lookup(v, self.api, self.endpoint)
v = self.default_ret(v, self.api, self.endpoint)

louis-oui:
[...]

PR #163, opened 2019-04-18
Update Endpoint #161

louis-oui:
Record url is not set + Endpoint is not updated #161
https://github.com/digitalocean/pynetbox/issues/161


louis-oui:
Update Endpoint #161

PR #179, opened 2019-09-25
Add :raises: AllocationError in Request.post() docstring

markkuleinio:
The :raises: description for AllocationError was missing from my earlier pull request. Hopefully the syntax is correct.


markkuleinio:
Unfortunately I don't know how to write a unit test for a case like this. But I'm willing to learn if you end up writing something...

Issue #165, opened 2019-04-25
If netbox is not reachable you just get JSONDecodeError


axlroden:
Python 3.7.2
netbox 2.5.10
pynetbox - latest

I have netbox running on port 8000
```
nb = pynetbox.api('http://127.0.0.1/', token='0123456789abcdef0123456789abcdef01234567')
nb.dcim.sites.all()
```
So this created traceback error, as the url is wrong..
Maybe have a more meaningful error message?

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/axlrod/.local/share/virtualenvs/gluestick-6gxqVldp/lib/python3.7/site-packages/pynetbox/core/endpoint.py", line 98, in all
    return [self._response_loader(i) for i in req.get()]
  File "/Users/axlrod/.local/share/virtualenvs/gluestick-6gxqVldp/lib/python3.7/site-packages/pynetbox/core/query.py", line 239, in get
    return req_all(self.url)
  File "/Users/axlrod/.local/share/virtualenvs/gluestick-6gxqVldp/lib/python3.7/site-packages/pynetbox/core/query.py", line 217, in req_all
    req = make_request(url)
  File "/Users/axlrod/.local/share/virtualenvs/gluestick-6gxqVldp/lib/python3.7/site-packages/pynetbox/core/query.py", line 212, in make_request
    return req.json()
  File "/Users/axlrod/.local/share/virtualenvs/gluestick-6gxqVldp/lib/python3.7/site-packages/requests/models.py", line 897, in json
    return complexjson.loads(self.text, **kwargs)
  File "/Users/axlrod/.pyenv/versions/3.7.2/lib/python3.7/json/__init__.py", line 348, in loads
    return _default_decoder.decode(s)
  File "/Users/axlrod/.pyenv/versions/3.7.2/lib/python3.7/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/Users/axlrod/.pyenv/versions/3.7.2/lib/python3.7/json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

PR #180, opened 2019-09-26
Import AllocationError from pynetbox.core.query, also add ContentError to handle json failures

markkuleinio:
This imports AllocationError from pynetbox.core.query to the package level in pynetbox/__init__.py.

### DIFF
--- a/pynetbox/__init__.py
+++ b/pynetbox/__init__.py
@@ -1,7 +1,6 @@
 from pkg_resources import get_distribution, DistributionNotFound
 
-from pynetbox.core.query import RequestError, AllocationError
-from pynetbox.core.query import ContentError
+from pynetbox.core.query import RequestError, AllocationError, ContentError
 from pynetbox.api import Api as api
 
 try:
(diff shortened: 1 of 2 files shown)


markkuleinio:
I see the system includes all the commits in the same pull request, no problem.

Anyway, by adding **ContentError** I'm trying to address #165 that is caused by the server outputting something unexpected, meaning that the server is not a NetBox server at all. `req.ok` is still `True`, so `req.json()` will fail. With ContentError:

```
>>> import pynetbox
>>> netbox = pynetbox.api("http://172.24.24.7/", token="That's a server redirecting to some page")
>>> try:
...     netbox.dcim.devices.all()
... except pynetbox.ContentError as e:
...     print(e)
...
The server returned invalid (non-json) data. Maybe not a NetBox server?
>>>
```

pynetbox/__init__.py
```diff
@@ -1,6 +1,7 @@
 from pkg_resources import get_distribution, DistributionNotFound
 
-from pynetbox.core.query import RequestError
+from pynetbox.core.query import RequestError, AllocationError
+from pynetbox.core.query import ContentError
```
zachmoody: I think this would be ok on L3 or did the linter complain?

markkuleinio: Oh, yes, no complaints now that I joined the lines. I just thought the limit was 72 characters! :-D

zachmoody: Nice. Thanks!

Issue #181, opened 2019-10-04
NetBox to deprecate _choices endpoints?


markkuleinio:
NetBox is discussing deprecating the `_choices` endpoints, and using Django OPTIONS requests instead: https://github.com/netbox-community/netbox/issues/3416

That would require changes in pynetbox as well.

Example output:

```
$ curl -H "Authorization: Token xxx" -X OPTIONS https://netbox.domain.local/api/dcim/interfaces/
 | jq .
{
  "name": "Interface",
  ...
  "actions": {
    "POST": {
      ...
      "type": {
        "type": "field",
        "required": false,
        "read_only": false,
        "label": "Type",
        "choices": [
          {
            "display_name": "Virtual",
            "value": 0
          },
          {
            "display_name": "OC-192/STM-64",
            "value": 6400
          },
          {
            "display_name": "CDMA",
            "value": 2820
          },
...
```

markkuleinio:
I'll close this as PR #183 was merged in, thanks!

PR #183, opened 2019-10-18
Add choices() in Endpoint using OPTIONS requests

markkuleinio:
As mentioned in #181, there is a discussion in NetBox (netbox-community/netbox#3416) to remove the _choices endpoints and use the Django-provided `OPTIONS` request.

This PR creates an options() method in Request class and a choices() method in Endpoint class, for using the `OPTIONS` interface. The existing App-level choices interface is still available for existing code to use it while supported by NetBox.

**Current choices style** (before this PR, with NetBox 2.6.6, lots of output omitted):

```
>>> from pprint import pprint
>>> pprint(netbox.ipam.choices())
...
 'ip-address:role': [{'label': 'Secondary', 'value': 20},
                     {'label': 'VIP', 'value': 40},
                     {'label': 'VRRP', 'value': 41},
                     {'label': 'Loopback', 'value': 10},
                     {'label': 'GLBP', 'value': 43},
                     {'label': 'CARP', 'value': 44},
                     {'label': 'HSRP', 'value': 42},
                     {'label': 'Anycast', 'value': 30}],
 'ip-address:status': [{'label': 'Active', 'value': 1},
                       {'label': 'Reserved', 'value': 2},
                       {'label': 'Deprecated', 'value': 3},
                       {'label': 'DHCP', 'value': 5}],
...
>>> pprint(netbox.dcim.choices())
...
 'device:face': [{'label': 'Front', 'value': 0}, {'label': 'Rear', 'value': 1}],
 'device:status': [{'label': 'Offline', 'value': 0},
                   {'label': 'Active', 'value': 1},
                   {'label': 'Planned', 'value': 2},
                   {'label': 'Staged', 'value': 3},
                   {'label': 'Failed', 'value': 4},
                   {'label': 'Inventory', 'value': 5},
                   {'label': 'Decommissioning', 'value': 6}],
...
 'site:status': [{'label': 'Active', 'value': 1},
                 {'label': 'Planned', 'value': 2},
                 {'label': 'Retired', 'value': 4}]}
...
```

**Examples with the new methods** (with NetBox 2.6.6):

```
>>> pprint(netbox.ipam.ip_addresses.choices())
{'role': [{'display_name': 'Secondary', 'value': 20},
          {'display_name': 'VIP', 'value': 40},
          {'display_name': 'VRRP', 'value': 41},
          {'display_name': 'Loopback', 'value': 10},
          {'display_name': 'GLBP', 'value': 43},
          {'display_name': 'CARP', 'value': 44},
          {'display_name': 'HSRP', 'value': 42},
          {'display_name': 'Anycast', 'value': 30}],
 'status': [{'display_name': 'Active', 'value': 1},
            {'display_name': 'Reserved', 'value': 2},
            {'display_name': 'Deprecated', 'value': 3},
            {'display_name': 'DHCP', 'value': 5}]}

>>> pprint(netbox.dcim.devices.choices())
{'face': [{'display_name': 'Front', 'value': 0},
          {'display_name': 'Rear', 'value': 1}],
 'status': [{'display_name': 'Offline', 'value': 0},
            {'display_name': 'Active', 'value': 1},
            {'display_name': 'Planned', 'value': 2},
            {'display_name': 'Staged', 'value': 3},
            {'display_name': 'Failed', 'value': 4},
            {'display_name': 'Inventory', 'value': 5},
            {'display_name': 'Decommissioning', 'value': 6}]}

>>> pprint(netbox.dcim.sites.choices())
{'status': [{'display_name': 'Active', 'value': 1},
            {'display_name': 'Planned', 'value': 2},
            {'display_name': 'Retired', 'value': 4}]}
```

### DIFF
--- a/pynetbox/core/endpoint.py
+++ b/pynetbox/core/endpoint.py
@@ -289,7 +289,7 @@ class Endpoint(object):
 
         :Returns: Dict containing the available choices.
 
-        :Example:
+        :Examples:
 
         >>> from pprint import pprint
         >>> pprint(netbox.ipam.ip_addresses.choices())
--- a/pynetbox/core/endpoint.py
+++ b/pynetbox/core/endpoint.py
@@ -52,6 +52,7 @@ class Endpoint(object):
             app=app.name,
             endpoint=name.replace("_", "-"),
         )
+        self._choices = None
 
     def _lookup_ret_obj(self, name, model):
         """Loads unique Response objects.
@@ -282,6 +283,46 @@ class Endpoint(object):
             return [self._response_loader(i) for i in req]
 
         return self._response_loader(req)
+
+    def choices(self):
+        """ Returns all choices from the endpoint.
+
+        :Returns: Dict containing the available choices.
+
+        :Example:
+
+        >>> from pprint import pprint
+        >>> pprint(netbox.ipam.ip_addresses.choices())
+        {'role': [{'display_name': 'Secondary', 'value': 20},
+                  {'display_name': 'VIP', 'value': 40},
+                  {'display_name': 'VRRP', 'value': 41},
+                  {'display_name': 'Loopback', 'value': 10},
+                  {'display_name': 'GLBP', 'value': 43},
+                  {'display_name': 'CARP', 'value': 44},
+                  {'display_name': 'HSRP', 'value': 42},
+                  {'display_name': 'Anycast', 'value': 30}],
+         'status': [{'display_name': 'Active', 'value': 1},
+                    {'display_name': 'Reserved', 'value': 2},
+                    {'display_name': 'Deprecated', 'value': 3},
+                    {'display_name': 'DHCP', 'value': 5}]}
+        >>>
+        """
+        if self._choices:
+            return self._choices
+
+        resp = Request(
+            base=self.url,
+            token=self.api.token,
+            private_key=self.api.private_key,
+            ssl_verify=self.api.ssl_verify,
+        ).options()
+        self._choices = {}
+        post_data = resp['actions']['POST']
+        for prop in post_data:
+            if 'choices' in post_data[prop]:
+                self._choices[prop] = post_data[prop]['choices']
+
+        return self._choices
 
 
 class DetailEndpoint(object):
--- a/pynetbox/core/endpoint.py
+++ b/pynetbox/core/endpoint.py
@@ -310,14 +310,21 @@ class Endpoint(object):
         if self._choices:
             return self._choices
 
-        resp = Request(
+        req = Request(
             base=self.url,
             token=self.api.token,
             private_key=self.api.private_key,
             ssl_verify=self.api.ssl_verify,
         ).options()
+        try:
+            post_data = req['actions']['POST']
+        except KeyError:
+            raise ValueError(
+                "Unexpected format in the OPTIONS response at {}".format(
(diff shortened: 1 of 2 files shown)


zachmoody:
Good stuff, nice work. Has this already landed in 2.6.6? I didn't see it mentioned in the release notes and the issue's still open.

markkuleinio:
Thanks! As I understand it, the `OPTIONS` interface is a feature of Django framework, and it is thus available "without extra work" in NetBox (and is available at least in NetBox 2.6.5+ as I've tested it). But I don't know about version dependency. Maybe best to have that clarified in the NetBox issue.

zachmoody:
Oh ok, I'm not getting much from a cursory curl, but I might need to dork around with the settings. We'll probably want to add some tests with this too. I'll see about getting some time to do that soon'ish if you don't want to take a pass at it.

markkuleinio:
It was told that it (`OPTIONS`) has "always" been available in DRF and in NetBox, apparently just noticed only recently.

I added a check for KeyError in the dict parsing, just to give a meaningful error message if `['actions']['POST']` is ever missing from the `OPTIONS` response (not sure if that is possible in practice but anyway), but I'm still not capable of writing unit tests for these additions... that's still some strange piece of art for me :-/

markkuleinio:
A couple of additional comments came in my mind. First, obviously the structure for fetching the choices is now different, as earlier it was (still is for now) App-level, but the new `choices()` is Endpoint-level. My thinking for now is that it is actually good because the content of the new choices is different (`display_name` instead of `label`). If desired, maybe there is a way to enumerate the endpoints under each app and then collect all the choices in the same response to mimic the current behaviour. That maybe means a differently named method (again, due to different content so cannot just replace old choices, so they will have to co-exist for some time).

Second, I didn't yet check the code for the old `choices()` but some error checking code is maybe needed to prevent unexpected errors when NetBox removes the `_choices` endpoint (unless the code already handles that situation gracefully; I'm on EEST so have to check it some other time:-D).

markkuleinio:
I see the current `choices()` uses the normal Request error handling, meaning that `RequestError` will be raised if the current `App.choices()` is attempted and `_choices` is not found in NetBox API at some point in the future.

markkuleinio:
If a `get_all_choices()` is desired (instead of the proposed App-specific `choices()`), it is apparently doable by first GETting all the apps via  `/api/`, then GETting all endpoints from each app, and finally requesting OPTIONS from each endpoint, saving and returning them as one dict.

On a second thought, how about just implementing a "`get_choice_value()`" right away in pynetbox? I mean, personally my only use cases are just that: I need a specific option value for the certain endpoint. Unless @zachmoody has objections, I think I will implement that in this PR...

zachmoody:
I'm a fan of where it's at in this PR, on the endpoint. The original idea was to put it on the endpoints, but it was more work :rofl: :sob: and made a small departure from NetBox's structure so :man_shrugging:. We'll deprecate `app.choices()` at some point. 

I also got around to figuring out why my OPTIONS method wasn't working. Seems you need to pass a token even with wide open authorization. Some mocked output should be all I need to put together some tests for this.

markkuleinio:
Ok, I'll then REST my case (pardon the pun), I'll keep `get_choice_value()` in my own application-specific library as it has always been. Feel free to merge the PR when appropriate.

Issue #166, opened 2019-05-01
Allow passing a timeout parameter at API at creation time


chaomodus:
It would be very useful if we could specify a timeout value that would be passed to `requests` operations as its timeout parameter, perhaps as an extra parameter to pynetbox.API().

zachmoody:
:wave:  you should be able to do this, though not as an init kwarg. Have some [examples in the docs](https://pynetbox.readthedocs.io/en/latest/advanced.html#timeouts).

PR #186, opened 2019-10-24
Add timeout parameter to API object

skatsaounis:
Solves #166

Signed-off-by: Stamatis Katsaounis <katsaouniss@gmail.com>

### DIFF
--- a/pynetbox/api.py
+++ b/pynetbox/api.py
@@ -129,6 +129,7 @@ class Api(object):
     :param str,optional private_key: Your private key.
     :param bool/str,optional ssl_verify: Specify SSL verification behavior
         see: requests_.
+    :param float, optional timeout: request timeout in seconds
     :raises ValueError: If *private_key* and *private_key_file* are both
         specified.
     :raises AttributeError: If app doesn't exist.
@@ -152,6 +153,7 @@ class Api(object):
         private_key=None,
         private_key_file=None,
         ssl_verify=True,
+        timeout=None,
     ):
         if private_key and private_key_file:
             raise ValueError(
@@ -176,7 +178,8 @@ class Api(object):
             token=token,
             private_key=private_key,
             ssl_verify=ssl_verify,
-            http_session=self.http_session
+            http_session=self.http_session,
+            timeout=timeout,
         )
         if self.token and self.private_key:
             self.session_key = req.get_session_key()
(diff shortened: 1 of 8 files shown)


janw:
Thanks you for this. It is long overdue.

skatsaounis:
Hi @janw I can see that there are conflicts with the current master branch. I will try to resolve them and update this PR. Plus, I will also take into account the comment on #187

skatsaounis:
Hi @zachmoody if you have a couple of minutes to spare please review this PR and let me know if it is valid or if it misses anything

skatsaounis:
@zachmoody Hi, if you think this PR is of no benefit, I would like to close it instead of re-resolving conflicts. Please let me know because I am doing housekeeping to my GitHub.

Kind regards,
Stamatis

zachmoody:
Hi @skatsaounis, thanks for the PR. There's definitely benefit to it, I'm just slightly hesitant to merge because I keep thinking (in the little time I do get to spend on this project) there's a more comprehensive approach we can take to expose these knobs in requests people need. If you don't mind keeping it open, we might still use it, but certainly don't feel like it's necessary to keep coming back and resolving conflicts. I'll take care of those if we merge.

skatsaounis:
Thank you for letting me know. There is no problem to leave this PR open, especially now that I am aware of what is happening. By the way, thanks for this extremely useful python client.

markkuleinio:
I'll toss my 0.05€ in... As I see it based on https://requests.readthedocs.io/en/master/_modules/requests/sessions/, the `timeout` value is not `Session`-specific but it needs to be specified separately for each request (which kind of makes sense). Thus, it seems we need to carry the timeout value separately to be used with each request within `pynetbox` calls, and giving it as an argument to `.Api()` (and saving it in the instance for further use) seems reasonable. Unless there is a vision that the timeout value should be different in different NetBox API calls of course...

zachmoody:
> **Edit:** And now I stumbled into https://hodovi.ch/blog/advanced-usage-python-requests-timeouts-retries-hooks/#setting-default-timeouts that uses adapters with `Session` to specify a default timeout value. So, if `Session` override is added (#229), users can also implement timeouts at their will.

:100:  Came across something similar in the FRs for requests. I think it's the way to go, but working through the best way to expose it. Initial feeling is that this may be a bit much to ask users to do in order to set something like timeouts, but maybe I'm wrong.

zachmoody:
This was implemented in v5.0. Closing out this PR.

PR #187, opened 2019-10-31
Add Api.api_version()

markkuleinio:
This PR adds an `api_version()` method to the `Api` object. Usage is straightforward:

```
>>> import pynetbox
>>> netbox = pynetbox.api("http://netbox.server.example/", token="xyz")
>>> netbox.api_version()
'2.6'
```

This prepares pynetbox for possible changes in NetBox API, like the one discussed in https://github.com/netbox-community/netbox/issues/3569 ("Replace numeric constants with slug values"). This PR enables a developer to prepare her/his code to handle different NetBox versions when the behaviour of NetBox is known to change from one version to another.

This would have been useful in 2.5->2.6 transition when the interface endpoint changed `form_factor` to `type`: the choices dict does not have any backwards compatibility. (`KeyError` exception handler was a practical way to handle that though.)

Version numbers can be compared at least like this:

```
>>> from pkg_resources import parse_version
>>> if parse_version(netbox.api_version()) < parse_version("2.7"):
...     print("Waiting for version 2.7")
...
Waiting for version 2.7
>>>
>>> if parse_version(netbox.api_version()) >= parse_version("2.6"):
...     print("Happy to use NetBox API 2.6+")
...
Happy to use NetBox API 2.6+
>>>

```

**Note:** There is a PR #186 submitted that adds `timeout` parameter in the `requests` calls. That needs to be added here as well if that is to be merged.


markkuleinio:
Interesting: in commit https://github.com/digitalocean/pynetbox/commit/50050eba038f3904868d30dd1dc075617df03628 (year ago) earlier code related to API versions was removed. That was not mentioned in the release list or change log so I didn't realize that earlier.

In any case, since NetBox evidently changes the compatibility from version to version, I see a need for being able to get the API version as needed.

pynetbox/api.py
```diff
@@ -16,6 +16,7 @@
 from pynetbox.core.endpoint import Endpoint
 from pynetbox.core.query import Request
 from pynetbox.models import dcim, ipam, virtualization, circuits
+import requests
```
zachmoody: I think the heavy lifting in this fits a bit better in `query.py`, plus we're already brining in `requests` there.

tests/test_api.py
```diff
@@ -87,3 +87,32 @@ def test_ssl_verify_false(self):
     def test_ssl_verify_string(self):
         kwargs = dict(def_kwargs, **{'ssl_verify': '/path/to/bundle'})
         self.common_arguments(kwargs, 'ssl_verify', '/path/to/bundle')
+
+
+class ApiVersionTestCase(unittest.TestCase):
```
zachmoody: Nice.

pynetbox/api.py
```diff
@@ -159,3 +160,30 @@ def __init__(
         self.tenancy = App(self, "tenancy")
         self.extras = App(self, "extras")
         self.virtualization = App(self, "virtualization")
+
+    def api_version(self):
```
zachmoody: Ultra minor nit, but I think just `version()` is a good name since we're already in the `api` object.

pynetbox/api.py
markkuleinio: Fair point. I created Request.get_version(), is that the way of doing it in query.py?

pynetbox/api.py
markkuleinio: Renamed that. Btw, should we also make `version` a property (with the decorator)?

zachmoody: :+1: Yeah, that's a good idea.

pynetbox/api.py
zachmoody: Yep, perfect.

markkuleinio:
I really need to get into habit of running `pycodestyle` myself before pushing :-D

markkuleinio:
I wonder why there is no sign of Travis doing its magics, while in https://travis-ci.org/digitalocean/pynetbox/pull_requests there is a success report for the last commit (46e1af5).

zachmoody:
> I really need to get into habit of running `pycodestyle` myself before pushing :-D

Sooner or later I think we'll switch to Black for formatting.

PR #199, opened 2019-11-16
Fix Endpoint.choices() by assigning http_session (to fix #198)

### LINKED ISSUE #198: Endpoint.choices() does not work in 4.1.0
`http_session` is not set in `.choices()` in version 4.1.0:

```
>>> netbox.dcim.devices.choices()
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.7/dist-packages/pynetbox/core/endpoint.py", line 327, in choices
    ssl_verify=self.api.ssl_verify,
  File "/usr/local/lib/python3.7/dist-packages/pynetbox/core/query.py", line 367, in options
    return self._make_call(verb="options")
  File "/usr/local/lib/python3.7/dist-packages/pynetbox/core/query.py", line 243, in _make_call
    req = getattr(self.http_session, verb)(
AttributeError: 'NoneType' object has no attribute 'options'
>>>

```

markkuleinio:
Fixes #198

### DIFF
--- a/pynetbox/api.py
+++ b/pynetbox/api.py
@@ -189,5 +189,6 @@ class Api(object):
         version = Request(
             base=self.base_url,
             ssl_verify=self.ssl_verify,
+            http_session=self.http_session,
         ).get_version()
         return version
--- a/pynetbox/core/endpoint.py
+++ b/pynetbox/core/endpoint.py
@@ -325,6 +325,7 @@ class Endpoint(object):
             token=self.api.token,
             private_key=self.api.private_key,
             ssl_verify=self.api.ssl_verify,
+            http_session=self.api.http_session,
         ).options()
         try:
             post_data = req['actions']['POST']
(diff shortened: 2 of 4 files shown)


markkuleinio:
Actually, why do we allow `http_session` default to `None` in `Request.__init__()`, when in `Request._make_call()` (which is used with every verb) there is:

```
        req = getattr(self.http_session, verb)(
            url_override or self.url, headers=headers, verify=self.ssl_verify,
            params=params, json=data
        )
```
--> if `http_session` is `None`, then an exception will be evident.

So actually `http_session` should **not** be an optional argument for `Request`.

markkuleinio:
I made also `base` a required argument in `Request.__init__()` because of
```
self.base = self.normalize_url(base)
```
(which will raise an exception if `base` is `None` or empty string). I would assume that if a `Request` object is instantiated then we do have an URL that we are dealing with... 😃

Issue #200, opened 2019-11-17
cannot access /api/extras/_custom_field_choices endpoint


brokenjacobs:
Callling:
```python
nb.extras._custom_field_choices.all()
```

Results in pynetbox sanitizing the _ to a - for some reason?
```
pynetbox.core.query.RequestError: The requested url: https://MYNETBOXHOST/api/extras/-custom-field-choices/ could not be found.
```

markkuleinio:
Looking at `api.py` I see there is no `App` method for `_custom_field_choices` so that needs to be written. I've never used custom field choices so maybe I'll have a look at that at some point and implement the method, unless @zachmoody has other views on this.

zachmoody:
Sounds like a good plan, unfortunate we have to add another special method just as we're talking about being able to get rid of `.choices()` :cry:. Thinking maybe a `.custom_choices()` method that's similar to `.choices()` but we raise `NotImplemented` if `self.name` is anything other than extras, what do you think?

markkuleinio:
Sounds good. I only now realized that `/_choices` also accept an `id`, like `prefix:status`... 😄 The `/_custom_field_choices` also does that, I will have to test the custom values first.

markkuleinio:
For the record, that's how the API output looks like for two custom fields, both having two different selectable options:

```
$ curl -H "Authorization: Token xxx" -H "Accept: application/json; indent=4" \
http://netbox.example.local/api/extras/_custom_field_choices/
{
    "Testfield": {
        "Test2": 2,
        "Test1": 1
    },
    "Testfield2": {
        "TF2": 4,
        "TF1": 3
    }
}

$ curl -H "Authorization: Token xxx" -H "Accept: application/json; indent=4" \
http://netbox.example.local/api/extras/_custom_field_choices/Testfield2/
{
    "TF2": 4,
    "TF1": 3
}
```

Looks like it could be possible to write a method like `custom_choices(self, field_name=None, value=None)` where the caller can optionally specify the `field_name` or `field_name` + `value` to get just the value that's needed.

PR #201, opened 2019-11-18
Add method to return custom choices

markkuleinio:
This implements a method for `/_custom_field_choices` for `extras` app (#200). Example:

```
>>> nb.extras.custom_choices()
{'Testfield': {'Test2': 2, 'Test1': 1}, 'Testfield2': {'TF2': 4, 'TF1': 3}}
```

Some thoughts:

If called with other app than `extras`, the standard `RequestError` will be raised (because of 404 error):

```
>>> nb.dcim.custom_choices()
...
pynetbox.core.query.RequestError: The requested url: http://netbox.example.local/api/dcim/_custom_field_choices/ could not be found.
```
If desired, specific trigger for `NotImplementedError` can be implemented (`if self.name != "extras" ...`), I just felt keeping this simple after all.

No caching is done (`self._custom_field_choices` etc) because then editing the custom field choices in NetBox would require restarting the application using `pynetbox` (or handling the cache flush some other way).

Some unit test is still needed.

### DIFF
--- a/pynetbox/api.py
+++ b/pynetbox/api.py
@@ -78,6 +78,24 @@ class App(object):
 
         return self._choices
 
+    def custom_choices(self):
+        """ Returns _custom_field_choices response from app
+
+        :Returns: Raw response from NetBox's _custom_field_choices endpoint.
+        :Raises: :py:class:`.RequestError` if called for an invalid endpoint.
+        """
+        custom_field_choices = Request(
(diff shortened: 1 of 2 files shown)


zachmoody:
:+1: I suppose a `RequestError` explains what's going on well enough. Just needs a test and we can merge.

PR #207, opened 2019-11-25
Skip params if url_override set in Request._make_call()

### LINKED ISSUE #206: Pagination infinite loop
I'm getting what seems to be an infinite loop of calls where the GET parameters are always appended.
For example calling `netbox.dcim.interfaces.filter(mgmt_only=True)` where `netbox` is an instance of `pynetbox.api`, those are the calls as logged with debug level:
```
GET /api/dcim/interfaces/?mgmt_only=True
GET /api/dcim/interfaces/?mgmt_only=True&limit=1345&offset=50
GET /api/dcim/interfaces/?limit=1000&mgmt_only=True&offset=1050&mgmt_only=True&limit=1345&offset=50
GET /api/dcim/interfaces/?limit=1000&mgmt_only=True&mgmt_only=True&offset=1050&mgmt_only=True&limit=1345&offset=50
GET /api/dcim/interfaces/?limit=1000&mgmt_only=True&mgmt_only=True&mgmt_only=True&offset=1050&mgmt_only=True&limit=1345&offset=50
GET /api/dcim/interfaces/?limit=1000&mgmt_only=True&mgmt_only=True&mgmt_only=True&mgmt_only=True&offset=1050&mgmt_only=True&limit=1345&offset=50
[...SNIP...]
```

For reference the current setting for `MAX_PAGE_SIZE` in the Netbox server is the default `1000`.

Calling instead the API directly for `GET /api/dcim/interfaces/?mgmt_only=True` I get as next page:
```
"next": "https://$HOSTNAME/api/dcim/interfaces/?limit=50&mgmt_only=True&offset=50",
```
And calling `GET /api/dcim/interfaces/?mgmt_only=True&limit=1345&offset=50` (pynetbox's second call) I get as next page:
```
"next": "https://$HOSTNAME/api/dcim/interfaces/?limit=1000&mgmt_only=True&offset=1050",
```
That would actually allow to correctly loop over the results, but pynetbox seems to be appending the previous parameters too.

Pynetbox version used: `4.2.0`

markkuleinio:
This seems to fix #206, but to be clear: this is a quick fix and I'm not sure at all if this causes some other problems somewhere, so if @zachmoody or someone else could check from some other viewpoint.

### DIFF
--- a/pynetbox/core/query.py
+++ b/pynetbox/core/query.py
@@ -235,10 +235,11 @@ class Request(object):
             headers["X-Session-Key"] = self.session_key
 
         params = {}
-        if self.filters:
-            params = self.filters
-        if add_params:
-            params.update(add_params)
+        if not url_override:
+            if self.filters:
+                params = self.filters
+            if add_params:
+                params.update(add_params)
 
         req = getattr(self.http_session, verb)(
             url_override or self.url, headers=headers, verify=self.ssl_verify,


markkuleinio:
Thinking out aloud. `url_override` is only used in `Request.get()`. According to the output by @volans- (in the original issue) the `next` item contains all the URL parameters from the first request, so with `url_override` set to next URL, `self.filters` can be ignored (= `params` is kept empty), so this should work.

zachmoody:
👍 I think this is a fine longterm solution. Thanks for the PR.

Issue #208, opened 2019-11-30
Fail to hash response records as 'Endpoint' object has no attribute 'name'


volans-:
I'm getting errors if trying to hash some records returned by the api. In my case I was adding them to a `set()`.
The error that I'm getting is:
```
    File pynetbox/core/response.py", line 218, in __hash__
        return hash((self.endpoint.name, self.id))
AttributeError: 'Endpoint' object has no attribute 'name'
```
And indeed if I manually get a record, let's say with:
```
api.dcim.devices.get(1)
```
Its `.endpoint` doesn't have a `name` property.
From a quick look at the `Endpoint` class implementation, it seems that the `name` passed to it doesn't get stored in the instance.
It seems to me that the issue was introduce in the recent #196 and that a quick way to fix it is to change it to: `self.endpoint.return_obj.__name__` if I understand the intention there.
The tests are passing because the whole `Endpoint` object is mocked and the `name` property manually set.

Tested with `pynetbox 4.2.1`.

fach:
A fix for this is in 4.2.2 which was released yesterday so you should be good to close this issue.

PR #210, opened 2019-12-10
Fix using Record objects in hashable collections

fach:
This PR fixes two dependent issues with Record objects now being hashable:

1. #208 reports  the `name` attribute missing from `Endpoint` objects. b76e09d7bcb11850acb91d1e6da9c4139672b8c6 now includes this as an instance attribute. There's some string (re)formatting of `name`, both in the constructor, which swap underscores for dashes, and in the `_lookup_ret_obj` method, which converts `name` to CamelCase. I went with the former, but open to feedback.

2. PR https://github.com/digitalocean/pynetbox/pull/196 introduced the ability to use Records in hashable collections. However, there is no `__eq__` method in the `Record` class defined, which causes two equal Record objects to be treated as unique since there is no implementation for comparison. This creates an issue when attempting to use Record objects in hashable collections, such as dictionary keys. The unit tests only check for the hash result being equal but do not exercise object comparison. b761930baf048f2d1bed9cfebd7e56e83d6bf194 adds a comparison operator as well as an additional unit test to exercise object comparison.

Repro code:
```
notify = {}                                                  
for c in circuits:                                           
    circuit = nb.circuits.circuits.get(cid=c)                
    provider = circuit.provider                              
    if provider in notify:                                   
        print(f'Duplicate provider {provider} found.')       
    else:                                                    
        print(f'New provider found {provider}. Inserting...')
        notify[provider] = c                                 
```

Without patch set:
```
~/code/notify# ./main.py
New provider found Telia. Inserting...
New provider found NTT. Inserting...
New provider found Telia. Inserting...
New provider found NTT. Inserting...
New provider found Telia. Inserting...
```
With patch set:
```
~/code/notify# ./main.py
New provider found Telia. Inserting...
New provider found NTT. Inserting...
Duplicate provider Telia found.
Duplicate provider NTT found.
Duplicate provider Telia found.
```

### DIFF
--- a/pynetbox/core/endpoint.py
+++ b/pynetbox/core/endpoint.py
@@ -42,6 +42,7 @@ class Endpoint(object):
 
     def __init__(self, api, app, name, model=None):
         self.return_obj = self._lookup_ret_obj(name, model)
+        self.name = name.replace("_", "-")
         self.api = api
         self.base_url = api.base_url
         self.token = api.token
@@ -50,7 +51,7 @@ class Endpoint(object):
         self.url = "{base_url}/{app}/{endpoint}".format(
             base_url=self.base_url,
             app=app.name,
-            endpoint=name.replace("_", "-"),
+            endpoint=self.name,
         )
         self._choices = None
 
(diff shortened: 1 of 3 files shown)


zachmoody:
It would help to use an attribute on Endpoint that actually exists, wouldn't it :man_facepalming:. Thanks for the PR!

Issue #212, opened 2020-01-04
Accessing "records" via attributes of related records does not update Endpoint


ChrisPortman:
If I .get() an item, e.g. an interface:
```
interface = client.dcim.interfaces.get(mac_address=mac)
```
and then access the device the interface belongs to via the interface's `device` attribute:
```
device = interface.device
```
and then try to alter the device and save:
```
device.display_name = "new.display.name"
device.save()
```
The PATCH to perform the save occurs against the interface's Endpoint:
```
[2020-01-04 20:39:39 +0000] [128] [DEBUG] PATCH /api/dcim/interfaces/1/
[04/Jan/2020:20:39:39 +0000] "PATCH /api/dcim/interfaces/1/ HTTP/1.0" 200 514 "-" "python-requests/2.22.0"
```
Looking at the code, it appears that if I were to do `device.delete()` it would actually delete the interface.

PR #213, opened 2020-01-05
Issue #212 Accessing related resources via parent attribute uses corr…

ChrisPortman:
…ect models

Fix an issue whereby retrieving an object directly, e.g:
```
interface = client.dcim.interfaces.get(1)
```
and then accessing a related object from that object:
```
device = interface.device
```
would result in device being of a generic Record() type and the Endpoint
being that of the original interface object.  This is dangerous because:
```
device.delete()
```
would actually call delete against the interface and not the device.

This change, parses the url value of related objects to infer the App()
and record model that is appropriate to generate a new record of the
correct type with its proper endpoint.

### DIFF
--- a/pynetbox/core/response.py
+++ b/pynetbox/core/response.py
@@ -16,7 +16,6 @@ limitations under the License.
 from pynetbox.core.query import Request
 from pynetbox.core.util import Hashabledict
 
-
 # List of fields that are lists but should be treated as sets.
 LIST_AS_SET = ("tags", "tagged_vlans")
 
@@ -240,6 +239,8 @@ class Record(object):
 
         def list_parser(list_item):
             if isinstance(list_item, dict):
+                if self.api and "url" in list_item:
+                    return self._parse_related_item(list_item)
                 return self.default_ret(list_item, self.api, self.endpoint)
             return list_item
 
@@ -252,6 +253,8 @@ class Record(object):
                     continue
                 if lookup:
                     v = lookup(v, self.api, self.endpoint)
+                elif self.api and "url" in v:
+                    v = self._parse_related_item(v)
                 else:
                     v = self.default_ret(v, self.api, self.endpoint)
                 self._add_cache((k, v))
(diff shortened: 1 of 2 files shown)


pynetbox/core/response.py
```diff
@@ -240,6 +239,8 @@ def _parse_values(self, values):
 
         def list_parser(list_item):
             if isinstance(list_item, dict):
+                if self.api and "url" in list_item:
+                    return self._parse_related_item(list_item)
```
ChrisPortman: check for self.api for the sake of unit tests.

pynetbox/core/response.py
```diff
@@ -252,6 +253,8 @@ def list_parser(list_item):
                     continue
                 if lookup:
                     v = lookup(v, self.api, self.endpoint)
+                elif self.api and "url" in v:
+                    v = self._parse_related_item(v)
```
ChrisPortman: check for self.api for the sake of unit tests.

pynetbox/core/response.py
```diff
@@ -265,6 +268,23 @@ def list_parser(list_item):
...
+        Returns:
+            Record object of the appropriate type
+            with the correct Endpoint,
+        """
+        url = values["url"]
+        app, name, _ = url.strip("/").split("/")[-3:]
+        app = getattr(self.api, app)
+        model = app.model
+
+        endpoint = self.endpoint.__class__(self.api, app, name, model)
```
ChrisPortman: We cant actually import the Endpoint class as the resource module is imported by the endpoint module resulting in a circular dependancy issue.  Hence self.endpoint.__class__ to access Endpoint()

pynetbox/core/response.py
```diff
@@ -265,6 +268,23 @@ def list_parser(list_item):
...
+
+        return endpoint.return_obj(values, self.api, endpoint)
```
ChrisPortman: Leverage the existing logic implemented in the endpoint to resolve model classes etc.

jsenecal:
I believe this could help in fixing https://github.com/digitalocean/pynetbox/issues/241 @zachmoody  if the conflicts are resolved.

zachmoody:
Think the issue with calling delete on nested objects was fixed in v4.2.4. Do you mind opening another issue for `interfaces > device` being a generic Record? We can fix that in the interface model.

PR #216, opened 2020-01-15
Add threading to .all() and .filter() requests

tyler-8:
~~**WIP** do not merge yet~~

~~Opening this PR just for tracking purposes.~~

Fixes: https://github.com/digitalocean/pynetbox/issues/191

### DIFF
--- a/pynetbox/core/query.py
+++ b/pynetbox/core/query.py
@@ -13,7 +13,10 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 """
-import concurrent.futures as cf
+try:
+    import concurrent.futures as cf
+except ImportError:
+    pass
 import json
 from six.moves.urllib.parse import urlencode
 
@@ -280,13 +283,8 @@ class Request(object):
                 )
 
             for future in cf.as_completed(futures_to_results):
-                try:
-                    result = future.result()
-                except Exception:
-                    # TODO: Raise exception with preferred method
-                    pass
-                else:
-                    ret.extend(result["results"])
+                result = future.result()
+                ret.extend(result["results"])
 
     def get(self, add_params=None):
         """Makes a GET request.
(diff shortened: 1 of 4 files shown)


tyler-8:
Build is failing because I'm using `concurrent.futures` which isn't available in Python 2. :\

tyler-8:
This was a function to grab all the VLANs.

```python
nb = pynetbox.api(url, token=token)
nb2 = pynetbox.api(url, token=token, threading=True)

no_threading = timeit(nb)
# Took 0:01:18.703560
len(no_threading)
# 51900

threading = timeit(nb2)
# Took 0:00:27.203155
len(threading)
# 51900
```

pynetbox/core/query.py
```diff
@@ -262,6 +270,24 @@ def _make_call(
...
+                new_params = {"offset": offset, "limit": page_size}
+                futures_to_results.append(
+                    pool.submit(self._make_call, add_params=new_params)
+                )
+
+            for future in cf.as_completed(futures_to_results):
+                try:
+                    result = future.result()
+                except Exception:
+                    # TODO: Raise exception with preferred method
```
tyler-8: I'd like to hear your thoughts on wait to do here @zachmoody - raise a `RequestError` and blow up the whole thing? Or perhaps a error log.

zachmoody: Admittedly haven't done a lot with thread, but what happens without the try-block?

tyler-8: Without the try block if the individual thread (one page of the API call) encountered an exception, that exception will be raised for the whole job.

zachmoody: Yeah, I think that's ok when the next best alternative (imo) is just catching and re-raising here. Kinda feel like getting an incomplete return should be terminal.

tyler-8: Done

tyler-8:
I think I've got everything wrapped up on this for now.

zachmoody:
Nice work! Will work on getting this merged and a new release cut today.

PR #222, opened 2020-02-08
Allowed return_fields to be passed into serialize method

### LINKED ISSUE #221: using strings for choice fields results in always updating an object
Python: 3.6.9
NetBox: 2.7.3
pynetbox: 4.2.4

NetBox 2.7.0 introduced **API Choice Fields Now Use String Values**:
https://github.com/netbox-community/netbox/releases/tag/v2.7.0

When using a string for a choice field, pynetbox always update the object, even if the data are not changed.

Here a test script:

```python
import os
import pynetbox

netbox = pynetbox.api(os.environ["NETBOX_URL"], token=os.environ["NETBOX_API_TOKEN"])
cluster = netbox.virtualization.clusters.get(name="test")
if cluster is None:
    cluster= netbox.virtualization.clusters.create(name="test", type=1)

data = dict(name="test", cluster=cluster.id, status="active")
vm = netbox.virtualization.virtual_machines.get(name=data["name"])
if vm is None:
    vm = netbox.virtualization.virtual_machines.create(data)
    print("vm '{}' created".format(data["name"]))

update = vm.update(data)
if update:
    print("vm '{}' updated".format(data["name"]))
```

Running this script multiple times always shows `vm 'test' updated`.
You see also a changelog entry for each run in NetBox.
But actually the virtual_machine data are always the same for each run, so I would not expect an update.

If you change `status="active"` to `status=1` in the script, pynetbox correctly not updates the object.

FragmentedPacket:
This allows the user to specify how they want the nested objects to be serialized.

This can fix #221, but that will inevitably be fixed once the API no longer returns IDs for choices.

This option does offer the user more control.

### DIFF
--- a/pynetbox/core/response.py
+++ b/pynetbox/core/response.py
@@ -302,7 +302,7 @@ class Record(object):
             return True
         return False
 
-    def serialize(self, nested=False, init=False):
+    def serialize(self, nested=False, init=False, return_fields=None):
         """Serializes an object
 
         Pulls all the attributes in an object and creates a dict that
@@ -320,7 +320,7 @@ class Record(object):
         :returns: dict.
         """
         if nested:
-            return get_return(self)
+            return get_return(self, return_fields=return_fields)
 
         if init:
             init_vals = dict(self._init_cache)
@@ -333,7 +333,8 @@ class Record(object):
             else:
                 if isinstance(current_val, Record):
                     current_val = getattr(current_val, "serialize")(
-                        nested=True
+                        nested=True,
+                        return_fields=return_fields
                     )
 
                 if isinstance(current_val, list):


zachmoody:
Hey @FragmentedPacket thanks for the PR. Does this work without being able to pass `return_fields` into `save()`? I'm thinking we can handle this without requiring a change in client code though. Can you see if the changes in PR #223 fix your issue?

FragmentedPacket:
@zachmoody Thanks for the fast response. PR #223 does fix my issue it appears and I'm find with going that route as to prevent any change in client code!

zachmoody:
Nice! Thanks for testing. Will get a new release out.

PR #227, opened 2020-03-20
Fix virtualization config context

JerradGit:
I noticed that the config_context field type was different between the device and virtualization models.

VM was
<class ‘pynetbox.core.response.Record’>

Device was
<class ‘dict’>


zachmoody:
Nice, thanks!

PR #232, opened 2020-04-13
Add deprecation note in App.choices()

markkuleinio:
NetBox 2.8.0 is out, and `App.choices()` does not work anymore. This PR adds a deprecation note in the method (and thus in the generated documentation as well), pointing the user to `Endpoint.choices()` instead. No changes in code.


zachmoody:
:+1:  thanks

PR #243, opened 2020-06-17
Fixes #241: Inconsistent hashing for nested api objects

### LINKED ISSUE #241: Inconsitent hashing for nested api objects
Let me try to explain this in a clear fashion:

When comparing the results of two api endpoints, for instance a `vlan` record and a specific vlan in the list of records returned in a specific `Interface` `tagged_vlans`, the hashing algorithm makes it so that even though the same vlan is represented, they don't have the same hash.

This is due to the fact that, for instance, `interface.tagged_vlans` nested list items have an `endpoint.name` of `interfaces` whilst the vlan record itself has an `endpoint.name` of `vlans`. Since the `endpoint.name` is used in the computation of the hash (see [pynetbox/core/response.py#L219-L231](https://github.com/digitalocean/pynetbox/blob/master/pynetbox/core/response.py#L219-L231)), this fails.

```
[...]
> netbox_interface.tagged_vlans[97].__class__
<class 'pynetbox.core.response.Record'>
> netbox_interface.tagged_vlans[97].vid
2164
> netbox_interface.tagged_vlans[97].id
856
> netbox_interface.tagged_vlans[97].endpoint.name
'interfaces'

> netbox_vlan.__class__
<class 'pynetbox.core.response.Record'>
> netbox_vlan.vid
2164
> netbox_vlan.id
856
> netbox_vlan.endpoint.name
'vlans'

> netbox_vlan == netbox_interface.tagged_vlans[97]
False
```

This also prevents things like `netbox_vlan in  netbox_interface.tagged_vlans` from working.

jsenecal:
### Fixes #241

As proposed here: https://github.com/digitalocean/pynetbox/issues/241#issuecomment-643430226 I went with option 1.

Let me know if you see a different fix for this.

### DIFF
--- a/pynetbox/core/response.py
+++ b/pynetbox/core/response.py
@@ -177,6 +177,7 @@ class Record(object):
                     r"(?P<app>\w*)/"
                     r"(?P<endpoint>\w*)/"
                     r"(?P<key>\w*)/"
+                    r"(\?(.*=.*&?)+)?"
                     r"(?!\w)$"
                 )
                 url_match = url_re.match(values['url'])
--- a/pynetbox/core/response.py
+++ b/pynetbox/core/response.py
@@ -13,6 +13,8 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 """
+import re
+
 from pynetbox.core.query import Request
 from pynetbox.core.util import Hashabledict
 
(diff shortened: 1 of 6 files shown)


zachmoody:
Hey @jsenecal, thanks for the PR. Going to merge this into a branch I've already got going with a slightly different approach to make the changes in PR #214 a little more elegant.

MattIPv4:
Hey, @jsenecal - Thanks a ton for this sweet PR! 😄

PR #247, opened 2020-06-24
sort lists for correct comparison later

cloos:
I sometimes noticed objects (like devices) are reporting True on .save(), but there where actually no changes.
I found out that the tags list of the initial and changed object sometimes don't have the same ordering, which is fixed by this PR.

I thought of adding a test like this to `tests/unit/test_response.py`:
```python
def test_serialize_tag_list(self):
    test_values1 = {"id": 123, "tags": ["foo", "bar"]}
    test1 = Record(test_values1, None, None).serialize()
    test_values2 = {"id": 123, "tags": ["bar", "foo"]}
    test2 = Record(test_values2, None, None).serialize()
    self.assertEqual(test1, test2)
```
But as this test will only fail sometimes, I think it makes no sense to add this test.
But maybe you have a better idea.

### DIFF
--- a/pynetbox/core/response.py
+++ b/pynetbox/core/response.py
@@ -346,6 +346,7 @@ class Record(object):
                     ]
                     if i in LIST_AS_SET:
                         current_val = list(set(current_val))
+                    current_val.sort()
                 ret[i] = current_val
         return ret
 


zachmoody:
Would like to dig more into why we're reordering lists. Have any more specifics around when it happens? Can it be consistently reproduced?

cloos:
The NetBox API doesn't return tags sorted:
```shell
curl -sX GET -H "Authorization: Token $NETBOX_API_TOKEN" -H "Accept: application/json; indent=2" "$NETBOX_URL/api/dcim/devices/?q=xxx" |jq '.results[0].tags'
[
  "san",
  "nfs",
  "gpfs"
]
```

Also users who update tags maybe doesn't have them sorted.

But to correctly detect a change in tags, you have to sort them both.

zachmoody:
We don't _have_ to sort to compare them. I think ideally we'd like to preserve the order from NetBox, but it doesn't seem that's happening at all. Suspect it's being reordered in conversion from set to list, but need to dig into it more.

cloos:
If you compare lists, you have to sort them:
```python
>>> list1 = ["foo", "bar"]
>>> list2 = ["bar", "foo"]
>>> list1 == list2
False
```

If you convert a list to a set, the items are `maybe` reordered:
```python
>>> list1 = ["foo", "bar"]
>>> set(list1)
{'bar', 'foo'}
```
I say `maybe` here because: https://docs.python.org/3.6/library/stdtypes.html#set-types-set-frozenset
> A set object is an unordered collection of distinct hashable objects.

So you can't rely on the ordering here.

pynetbox is really great in only updating objects in NetBox if they have changes.
But this doesn't work for tags, which this PR fixes for me.

Just add the test I mentioned above and run the tests multiple times.
You will see that it sometimes fails.
With the change from this PR, this test doesn't fail anymore.

zachmoody:
> If you compare lists, you have to sort them

Yeah, comparing lists considers order, but it doesn't necessitate sorting. Not that I think it's common, or sane, but the order itself can convey meaning that'd be lost if we took it upon ourselves to sort. This may all be pointless if order from NetBox is inconsistent, but I pushed a PR that should fix the issue and preserve order we receive in #271. Thanks for reporting the issue and the PR.

PR #263, opened 2020-07-05
Fixes #262: Missing or duplicated results for .filter() when threading is enabled

### LINKED ISSUE #262: Missing or duplicated devices when threading is enabled
While trying to improve the time it takes to pull info about 83 devices, I enabled threading (introduced in PR 216) and lowered the MAX_PAGE_SIZE to 20. It saved me a few seconds, though the results became inconsistent.

```python
#!/usr/bin/env python
import pynetbox
nb = pynetbox.api(
    "http://URL",
    token="TOKEN",
    threading=True,
)
print(len(nb.dcim.devices.filter(role=["switch", "router"], has_primary_ip=True)))
```
```bash
(venv) $ time ./temp_netbox.py
83
./temp_netbox.py  0,24s user 0,04s system 18% cpu 1,512 total
(venv) $ time ./temp_netbox.py
66
[...]
./temp_netbox.py  0,23s user 0,04s system 14% cpu 1,879 total
(venv) $ time ./temp_netbox.py
83
./temp_netbox.py  0,21s user 0,01s system 9% cpu 2,385 total
(venv) $ time ./temp_netbox.py
49
./temp_netbox.py  0,29s user 0,03s system 21% cpu 1,455 total
```

In the actual data I could see either a lot of duplicates or just missing devices.
The weird part is that sometimes it's every other request that gets affected, but could also be like dozens of normal results in a row.
Environment:
* pynetbox 4.3.3 / 5.0.1;
* CPython 3.8.2 on Ubuntu 20.04 / Nginx Unit 1.18 + python:3.7.7-slim-buster Docker image;
* netboxcommunity/netbox:v2.8.6-ldap / netboxcommunity/netbox:v2.8.7-ldap.

One more thing: I haven't seen the issue when calling nb.dcim.devices.all() (also with threading enabled and MAX_PAGE_SIZE of 20), so I currently assume it happens only with the nb.dcim.devices.filter().

Please, let me know if there's anything I could help you with to narrow it down.

UPD: I guess I should start with taking a traffic dump. Any advices on the Python side are appreciated :)

weisdd:
As mentioned in #262, we sometimes see missing or duplicated results when calling nb.dcim.devices.filter() with threading enabled.
Troubleshooting showed that the requests themselves are scheduled with correct offsets, though when the actual calls happen, params seem to be overwritten, further causing incorrect results. Not always, but often enough, and especially visible with low MAX_PAGE_SIZE values.
image

In the implementation of query.py, I've noticed that concurrent sessions seem to use the same copy of the attrubute "self.filters" (that params = self.filters). That's why params.update(add_params) would sometimes cause collisions depending on the time when a particular request is called.
```python
        params = {}
         if not url_override:
             if self.filters:
                 params = self.filters
             if add_params:
                 params.update(add_params)
```
An example of this dict:
```python
{'role': ['switch', 'router'], 'has_primary_ip': True, 'limit': 20, 'offset': 60}
```
I think changing it to: 
```python
        params = {}
        if not url_override:
            if self.filters:
                params.update(self.filters)
            if add_params:
                params.update(add_params)
```
will help to mitigate the issue. At least, in my tests it seems to work.

UPD: a bit rewritten my text to make it clearer.

### DIFF
--- a/pynetbox/core/query.py
+++ b/pynetbox/core/query.py
@@ -250,7 +250,7 @@ class Request(object):
         params = {}
         if not url_override:
             if self.filters:
-                params = self.filters
+                params.update(self.filters)
             if add_params:
                 params.update(add_params)
 


weisdd:
@zachmoody :)

zachmoody:
Makes sense, nice catch.

tyler-8:
Sorry for the headache from my introduction of threading, great find!

weisdd:
@tyler-8 No worries, after working in TAC, I have a soft spot for troubleshooting activities :)
And thanks for the threads, in our use case they make a huge difference!

MattIPv4:
Hey, @weisdd - Thanks a ton for this sweet PR! 😄

Issue #242, opened 2020-06-16
Problem with update local_config_context on an existing device


nazarko-es:
Hello, i have problem with update local_config_context on an existing device.
My environment
Python version: 3.8
NetBox version: 2.8.4

Steps to Reproduce
Upload device with local_context_data field. For example my field:
{
"bgp_neighbors": [
{
"as": "65000",
"neighbor": "169.254.0.1"
},
{
"as": "65001",
"neighbor": "169.254.0.2"
}
],
"track": [
"10.10.0.251/32",
"10.10.0.252/32"
],
"vlans": []
}
Get local_context_data in python script and append local_context_data['vlans']. For example:
device = nb.dcim.devices.get(name='mydevice')
local_context = device.local_context_data
vlans_list = [{'vid': 1, 'name': 'vlan1'}, {'vid': 2, 'name': 'vlan2'}]
for vlan in vlans_list:
local_context['vlans'].append({"vid": vlan['vid'], 'name': vlan['name']})
device.update(dict(local_context_data=local_context))

device.update doesn't work, but if i don't use "get" from netbox, but set local_context in my code, it's work. For example:

device = nb.dcim.devices.get(name='pd100-elf-025')
new_local_context = {
"bgp_neighbors": [
{
"as": "65000",
"neighbor": "169.254.0.1"
},
{
"as": "65001",
"neighbor": "169.254.0.2"
}
],
"track": [
"10.10.0.251/32",
"10.10.0.252/32"
],
"vlans": []
}

vlans_list = [{'vid': 1, 'name': 'vlan1'}, {'vid': 2, 'name': 'vlan2'}]
for vlan in vlans_list:
new_local_context['vlans'].append({"vid": vlan['vid'], 'name': vlan['name']})
device.update(dict(local_context_data=new_local_context))

I tried it in postman:
If request metod is "put" - local_config_context will be replace
If request metod is "patch" - local_config_context will be update

markkuleinio:
Your `update()` syntax looks strange. For me it works just fine like this:

```
>>> d = netbox.dcim.devices.get(...)
>>> d.local_context_data
{'vlans': []}
>>> local_context
{'vlans': [{'vid': 1, 'name': 'vlan1'}, {'vid': 2, 'name': 'vlan2'}]}
>>> d.update({"local_context_data": local_context})
True
>>> d.local_context_data
[...]

PR #266, opened 2020-07-07
Fixes 242: Copy local_context_data as-is in Record._add_cache()

markkuleinio:
In #242 the problem seems to be that `local_context_data` contents are not properly copied to `Record._init_cache`, so `update()` and `save()` operations fail to recognize the changed contents of `local_context_data`.

This PR adds a check that for `local_context_data` it uses `copy.deepcopy()` to copy the dict contents in the init cache (and all other fields are handled as earlier). `local_context_data` only contains user-entered data so no specific attribute checks should be performed for it.

With this change the test code in https://github.com/digitalocean/pynetbox/issues/242#issuecomment-654988849 works and saves the `local_context_data` change.

@zachmoody feel free to evaluate if some other way to resolve the issue is more correct. Btw, there is an unused `Record._compare()` method but I didn't touch it yet. I can remove it as well if needed. (Update: oops `_compare()` seems to be used in tests I see)

### DIFF
--- a/pynetbox/core/response.py
+++ b/pynetbox/core/response.py
@@ -13,6 +13,8 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 """
+import copy
+
 import pynetbox.core.app
 from six.moves.urllib.parse import urlsplit
 from pynetbox.core.query import Request
@@ -235,7 +237,10 @@ class Record(object):
 
     def _add_cache(self, item):
         key, value = item
-        self._init_cache.append((key, get_return(value)))
+        if key == "local_context_data":
+            self._init_cache.append((key, copy.deepcopy(value)))
+        else:
+            self._init_cache.append((key, get_return(value)))
 
     def _parse_values(self, values):
         """ Parses values init arg.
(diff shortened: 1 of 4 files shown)


zachmoody:
> Btw, there is an unused Record._compare() method but I didn't touch it yet. I can remove it as well if needed. (Update: oops _compare() seems to be used in tests I see)

Yeah, does seem to be unused. Removing it probably wouldn't be an issue.

I like the approach. Might be a good idea to add a test for this in `test_response` for the `_diff` method. Just so we don't get any surprising behavior though.

markkuleinio:
I added a change for `local_context_data` field in the existing `test_diff()`, and also changed all the tests that used `_compare()` instead of `_diff()`.

markkuleinio:
I tried to change the test to catch the situation prior to `_add_cache()` change and noticed that `Record._parse_values()` did not copy the `local_context_data` at all, so I changed `_parse_values()` as well. I'm still slightly unsure if I'm doing the correct things so let's see 😃

zachmoody:
That's strange it didn't even add the attribute as a record object. Makes sense you'd have to update `_parse_values()` as well though. This looks good.

Sorry the tests are mess. I'm (slowly) trying to get to the point where we're mostly unit-testing everything and the test_blah files only do the generic integration tests. They're a bit overloaded right now as you found with the `_compare()` tests. Would've been totally fine to just remove those instead, but not going to complain about a net increase in testing. :grin:

markkuleinio:
Thanks, I missed one thing though: the # from the issue number in the title 😂

zachmoody:
haha, no worries, closed it out.

Issue #270, opened 2020-07-15
Pagination in get()


Dees7:
Hello. 
I have a lot of device and if I do dcim.devices.all() it take very long time. 
I'd like to work with devices one by one. But I can't  find how to do this. In API we have limit and offset.
Is it possible using pagination with pynetbox 

pynetbox (4.3.1)

tyler-8:
You could do something like this:

```python
devices = nb.dcim.devices.filter(brief=1)  # Very fast retrieval of device data

for device in devices:
    device.full_details()
    # Do things with the device Record object
```

zachmoody:
Should be fixed in v6, `.all()` is a generator now you can step through pages as you iterate through results.

PR #273, opened 2020-07-17
Allow iteration for result from Endpoints

sedyvlk:
endpoint's `all()` and `filter()` return a list. It makes take minutes to get all records and especially when there are a lot of objects, it is useful to start processing the data as it arrives 
This also partially implements #270 - not entirely as it does not paginate but returns the objects as it gets them #

### DIFF
--- a/pynetbox/core/endpoint.py
+++ b/pynetbox/core/endpoint.py
@@ -75,7 +75,7 @@ class Endpoint(object):
     def _response_loader(self, values):
         return self.return_obj(values, self.api, self)
 
-    def iall(self):
+    def iterall(self):
         """Queries the 'ListView' of a given endpoint.
 
         Returns all objects from an endpoint.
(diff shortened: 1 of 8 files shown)


mngan:
This doesn't really solve the problem. When the  req.get() call is made in the all() method, it will fetch all the records and return them as a list. This is ultimately what needs to be converted into an iterator, so that the paging requests are done iteratively.

sedyvlk:
I agree with you that it would be ideal to make the change in `.all()`, however, at the moment it returns a list. I certainly can replace `.all()` (and `.filter()` should not be forgotten) with an iterator but I will also break someone else's code if they index into returned data. In this instance, I prefer the stability of pynetbox API. So at the moment, you get two new methods. New method approach seems less hacky than adding a parameter to `.all()` that could turn it into a generator function.

If you have any other proposals?  Very happy to look at it

zachmoody:
In addition to @mngan's point that more than just Endpoint methods need to be turned into generators (I'm thinking `Request._make_call()` ultimately needs to be); I think it makes more sense to turn `.all()` and `.filter()` into generators. We'll need to cut a new major release, which I probably don't want to do immediately, but it gives us some time to get the implementation right.

zachmoody:
Closing this out as it was recently implemented in v6

PR #274, opened 2020-07-22
Fixes paginated get requests

mngan:
Fix limit in req_all. Should be the len of records returned not the total number of records
Fix req_all_threaded. Setting the limit to zero will fetch all the records, not the default page size

### DIFF
--- a/pynetbox/core/query.py
+++ b/pynetbox/core/query.py
@@ -311,7 +311,7 @@ class Request(object):
                     if first_run:
                         req = self._make_call(
                             add_params={
-                                "limit": req["count"],
+                                "limit": len(req["results"]),
                                 "offset": len(req["results"]),
                             }
                         )
@@ -324,9 +324,6 @@ class Request(object):
                 return req
 
         def req_all_threaded(add_params):
-            if add_params is None:
-                # Limit must be 0 to discover the max page size
-                add_params = {"limit": 0}
             req = self._make_call(add_params=add_params)
             if isinstance(req, dict) and req.get("results") is not None:
                 ret = req["results"]


pynetbox/core/query.py
```diff
@@ -311,7 +311,7 @@ def req_all():
                     if first_run:
                         req = self._make_call(
                             add_params={
-                                "limit": req["count"],
+                                "limit": len(req["results"]),
```
tyler-8, line 314: This is functionally doing the same thing. `len(req["results"])` and `req["count"]` would be the same number.

pynetbox/core/query.py
```diff
@@ -324,9 +324,6 @@ def req_all():
                 return req
 
         def req_all_threaded(add_params):
-            if add_params is None:
-                # Limit must be 0 to discover the max page size
-                add_params = {"limit": 0}
```
tyler-8, line 329: Removing this isn't ideal IMO. The default settings in NetBox for pagination are `MAX_PAGE_SIZE = 1000` and `PAGINATE_COUNT = 50` (the default `limit` if none is specified). If you specify `{"limit": 0}` in your query parameters, you will get at maximum `MAX_PAGE_SIZE` (1000 if using default) records.

The only way this behavior changes is if a NetBox admin sets their `MAX_PAGE_SIZE` to 0, which disables the limit entirely - and is a terrible idea that no one should be doing - which is why it's called out in the `pynetbox` [README](https://github.com/digitalocean/pynetbox/blob/72e62aa831e594753f232f17a6c12c4d8bbca03c/README.md#threading).

So with all that in mind, if the `{"limit": 0}` is removed, then `pynetbox` will be stuck paginating at the `PAGINATE_COUNT ` size - which is quite low, at the default of 50, for most use cases. The reason that `{"limit": 0}` is used is so we can determine the largest page size available to speed up the rest of the calls.

I think a better alternative to doing this would be to provide an optional parameter for allowing the user to specify their own limit/page size.

zachmoody: > I think a better alternative to doing this would be to provide an optional parameter for allowing the user to specify their own limit/page size.

Agreed, let's rally on #273 to get that added.

PR #278, opened 2020-07-25
Function getnext() to fetch one by one

Dees7:
Function getnext() returns 1 item to fetch one by one.

Returns the first or selected N-th element. Return None if N more them items.
For example:
````
i=0
device=nb.dcim.devices.getnext(i)
while device:
 print(device)
 device=nb.dcim.devices.getnext(i)
 i=i+1
````

### DIFF
--- a/pynetbox/core/endpoint.py
+++ b/pynetbox/core/endpoint.py
@@ -100,6 +100,23 @@ class Endpoint(object):
 
         return response_loader(req.get(), self.return_obj, self)
 
+    def getnext(self, *args):
+
+        try:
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by sedyvlk** [comment]
for my benefit could you please elaborate on what are you trying to change/fix?

As the moment you can already do 

```
for device in nb.dcim.devices.all():
    ...
```

Dees7:
nb.dcim.devices.all() works very long time. Especially nb.ipam.ip_addresses.all(). 
Cause we have a lot of devices and IPs.

PR #279, opened 2020-08-01
Fixed req_all to actually return all pages of paginated results

craized:
Since the first "next" request will always be on first_run, the previous code was preventing using a url_override and only returning the first page of results under all circumstances.

First code contribution here, my apologies if I should have gone through different channels first. I'm happy to redirect if required. I think this is just a fix to a small oversight, otherwise I'm not understanding the desired behavior here.

### DIFF
--- a/pynetbox/core/query.py
+++ b/pynetbox/core/query.py
@@ -303,21 +303,8 @@ class Request(object):
             req = self._make_call(add_params=add_params)
             if isinstance(req, dict) and req.get("results") is not None:
                 ret = req["results"]
-                first_run = True
                 while req["next"]:
-                    # Not worrying about making sure add_params kwargs is
-                    # passed in here because results from detail routes aren't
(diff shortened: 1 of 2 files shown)


craized:
Ok, well I definitely should have run tests before submitting this. My second commit updates the test to match the actual results. In doing so, I removed the 'limit' and 'offset' params, but by my understanding, sending those to .all() is not supported anyway.

Cheers

zachmoody:
Yeah, maybe open an issue with what you're trying to do and what's not working. This is pretty integral stuff the normal operation of pynetbox.

craized:
Jeez, well I'm going to chalk this all up to frustrated Friday-brain, chasing the wrong problem since the behavior I'm seeing today is unrelated.

Thanks for reviewing anyway, I'll post an issue next time

zachmoody:
:)

PR #280, opened 2020-08-01
Added a trace property to the traceable interface classes

raddessi:
I have this in a working state here, but the code is not solid enough I would feel happy about merging it. More I would like your feedback on how you would like to approach this. The main issue is the `response_loader` function needing to be able to handle multiple `return_obj` types since the trace endpoint returns multiple types of items from the api - cables, interfaces, front ports, console ports, console server ports, etc.

I saw some places where you already inspected the uri to determine object type so that is what I've done here.. I'm not married to it though. Thanks for all the work on this project! I love the module, it's been a huge help :)


raddessi:
I would also like to add this for the other items that have `trace` methods, but most of those will need to be created in the dcim submodule first so I have not done that yet, pending your thoughts on how to proceed.

pynetbox/core/endpoint.py
```diff
@@ -21,7 +21,36 @@
...
+                            next(
+                                (
+                                    return_obj_class(ii, endpoint.api, endpoint)
+                                    for (return_obj_uri, return_obj_class) in return_obj.items()
+                                    if return_obj_uri in ii["url"]
+                                )
+                            )
+                        )
+                    else:
+                        # the last trace can consist of [cable_a, None, None] if there is no
```
raddessi, line 47: copy pasta, I will remove

zachmoody:
Interesting, thanks for the PR. I haven't really dug into the trace API yet. I feel like it's unique enough that I'd probably like to deal with it in the `.trace` property instead of messing with the more universal `response_loader`. We could call RODetailEndpoint and then return what we needed to from its return.

raddessi:
Ah ok that makes good sense. I'll play around with it and send something over. Thanks!

pynetbox/models/dcim.py
```diff
@@ -13,12 +13,56 @@
...
+from six.moves.urllib.parse import urlsplit
+
+from pynetbox.core.query import Request
 from pynetbox.core.response import Record, JsonField
 from pynetbox.core.endpoint import RODetailEndpoint
 from pynetbox.models.ipam import IpAddresses
 from pynetbox.models.circuits import Circuits
 
 
+class TraceableRecord(Record):
```
raddessi, line 25: I don't love having this here but the trace fn needs to use the dcim objects in the response.. if you have a better way to structure it let me know :)

pynetbox/models/dcim.py
```diff
@@ -13,12 +13,56 @@
...
+                if url_path.startswith("/api/dcim/cables"):
+                    return_obj_class = Cables
+                elif url_path.startswith("/api/dcim/front-ports"):
+                    return_obj_class = FrontPorts
+                elif url_path.startswith("/api/dcim/interfaces"):
+                    return_obj_class = Interfaces
+                elif url_path.startswith("/api/dcim/rear-ports"):
+                    return_obj_class = RearPorts
+                else:
+                    raise NotImplementedError(
```
raddessi, line 54: is this how you would want to handle this situation?

pynetbox/models/dcim.py
```diff
@@ -154,7 +222,21 @@ def __str__(self):
 
 class Cables(Record):
     def __str__(self):
-        return "{} <> {}".format(self.termination_a, self.termination_b)
+        # populate the terminations to get the full names if they are not already
```
raddessi, line 225: this is more of a bugfix/enhancement, it's related but I can split it out to a dedicated PR if you want

raddessi:
re: the failing tests.. yeah I have never seen python complain that `self` actually needs to be passed back in to a function when calling it but I got this error when testing locally and that seemed to fix it in my case: 
```
~/dev/pynetbox/pynetbox/core/response.py in __repr__(self)
    215 
    216     def __repr__(self):
--> 217         return str(self)
    218 
    219     def __getstate__(self):

~/dev/pynetbox/pynetbox/models/dcim.py in __str__(self)
    227             termination_a_name = self.termination_a.name
    228         except AttributeError:
--> 229             self.termination_a.full_details()
    230             termination_a_name = self.termination_a.name
    231 

TypeError: full_details() missing 1 required positional argument: 'self'
```

I'll dig in to why it's wanting that locally and fix that issue. Comments would still be appreciated, thanks!

pynetbox/models/dcim.py
```diff
@@ -154,7 +222,26 @@ def __str__(self):
...
+        try:
+            termination_a_name = self.termination_a.name
+        except AttributeError:
+            try:
+                self.termination_a.full_details()
+            except TypeError:
+                self.termination_a.full_details(self)
+            termination_a_name = getattr(self.termination_a, "name", self.termination_a)
```
raddessi, line 233: I don't really like this logic but I'm not sure of a better method. Ideas welcomed

raddessi:
@zachmoody do you have any thoughts?

zachmoody:
> @zachmoody do you have any thoughts?

Yeah, been meaning to review it, sorry to leave you hanging. Will do my best to get to it this week.

raddessi:
No worries, this is working as needed for my current work just fine s as long as it's still on your radar I'm happy. Thanks :)

pynetbox/models/dcim.py
zachmoody: Yeah, that's fair, I don't immediately see any better ways around it if we want to return objects. Will keep thinking about it though.

pynetbox/models/dcim.py
zachmoody: I'd probably make a dict here that mapped the endpoint name it might encounter with the custom object instead of an if/else tree (really wish switches were a thing in python). e.g.
```
{
    "cables": Cables,
    "front-ports": FrontPorts,
...
}
```
You'll probably want to add some other endpoint/objects you might come across in the traces as well like ConsolePort/ConsoleServerPort and PowerPort/Outlets. I'd probably default to just a simple `Record` object at the end of it instead of raising.

pynetbox/models/dcim.py
```diff
@@ -13,12 +13,56 @@
...
+                elif url_path.startswith("/api/dcim/interfaces"):
+                    return_obj_class = Interfaces
+                elif url_path.startswith("/api/dcim/rear-ports"):
+                    return_obj_class = RearPorts
+                else:
+                    raise NotImplementedError(
+                        "unable to unpack item data from endpoint '{}'".format(url_path)
+                    )
+                this_hop_ret.append(
+                    return_obj_class(hop_item_data, self.endpoint.api, self.endpoint)
```
zachmoody, line 58: You _should_ be able to just supply the `hop_item_data` and it'll figure out the rest of what it needs to based on the value of `url`.

tests/test_dcim.py
```diff
@@ -526,11 +526,31 @@ def test_get_circuit(self):
...
+                    ),
+                    call(
+                        "http://localhost:8000/api/{}/{}/1/".format(
+                            "circuits", "circuit-terminations"
+                        ),
+                        headers=HEADERS,
+                        params={},
+                        json=None,
+                    ),
+                ]
```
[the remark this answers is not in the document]
zachmoody: Yeah, have a look at the previous comment, the `__str__()` method might have something to do with it. Either way, we'd definitely want some tests that are specifically testing `.trace`.

zachmoody (changes requested):
Great progress, made some comments below. Was thinking about maybe changing this from a property to a method, it's a bit of a nit but you only get one shot at getting that part right without cutting a new release :smile:.

pynetbox/models/dcim.py
raddessi: Ahh.. nice, I'll try that.

pynetbox/models/dcim.py
raddessi: This was a change to force each of the returned items along the trace to get fully populated so that when you printed their name in the trace it wasn't just displayed as a `Record` item with no data. Once each item was accessed they would then print their name properly, but having a trace full of unlabeled hops was not very useful. It's not very clear, and I don't like it at all. I think it may be better to do something like this in the trace function itself (I think maybe you mentioned that before?) so that only on a trace will it try to prefetch the data of each item in the hops so they print out nicely.

tests/test_dcim.py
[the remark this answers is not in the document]
raddessi: Yeah. I'll also add testing with the next round of changes.

raddessi (commented):
Thanks for all the great feedback. I'm sorry I dropped off the map, life got in the way. 

> Was thinking about maybe changing this from a property to a method, it's a bit of a nit but you only get one shot at getting that part right without cutting a new release

I'll definitely change this from a property to a method in the next update as well unless I hear back from you, I think that's probably better anyway but I was trying to match styles.

zachmoody:
> Thanks for all the great feedback. I'm sorry I dropped off the map, life got in the way.

Ha, no worries, been in the same boat myself.

pynetbox/models/dcim.py
zachmoody: That's weird. Do you have an example you can show of the old behavior? Those should be Termination objects, but either way even Record's default __str__ behavior is to return .name, iirc.

pynetbox/models/dcim.py
raddessi: Sure, with this section of code reverted this is what I'm getting now:

```
In [1]: iface.trace()                                                                                                             
Out[1]: 
[[em1,
  <class 'pynetbox.models.dcim.Termination'> <> <class 'pynetbox.models.dcim.Termination'>,
  pair-11 (ports 21-22)],
 [port-2,
  <class 'pynetbox.models.dcim.Termination'> <> <class 'pynetbox.models.dcim.Termination'>,
  port-2],
 [pair-11 (ports 21-22),
  <class 'pynetbox.models.dcim.Termination'> <> <class 'pynetbox.models.dcim.Termination'>,
  Ethernet11]]

```

So it looks like maybe I was mistaken or confused about only the middle (not the first or last) having the issue, it may be all of the items returned from the trace until they are accessed in some way. You can see here how the info is filled out once the item is accessed directly: 
```
In [2]: trace = iface.trace()                                                                                                     

In [3]: trace                                                                                                                     
Out[3]: 
[[em1,
  <class 'pynetbox.models.dcim.Termination'> <> <class 'pynetbox.models.dcim.Termination'>,
  pair-11 (ports 21-22)],
 [port-2,
  <class 'pynetbox.models.dcim.Termination'> <> <class 'pynetbox.models.dcim.Termination'>,
  port-2],
 [pair-11 (ports 21-22),
  <class 'pynetbox.models.dcim.Termination'> <> <class 'pynetbox.models.dcim.Termination'>,
  Ethernet11]]

In [4]: trace[0][1]                                                                                                               
Out[4]: em1 <> pair-11 (ports 21-22)

In [5]: trace                                                                                                                     
Out[5]: 
[[em1, em1 <> pair-11 (ports 21-22), pair-11 (ports 21-22)],
 [port-2,
  <class 'pynetbox.models.dcim.Termination'> <> <class 'pynetbox.models.dcim.Termination'>,
  port-2],
 [pair-11 (ports 21-22),
  <class 'pynetbox.models.dcim.Termination'> <> <class 'pynetbox.models.dcim.Termination'>,
  Ethernet11]]

```

pynetbox/models/dcim.py
raddessi: Sure, with this section of code reverted this is what I'm getting now:

```
In [1]: iface.trace()                                                                                                             
Out[1]: 
[[em1,
  <class 'pynetbox.models.dcim.Termination'> <> <class 'pynetbox.models.dcim.Termination'>,
  pair-11 (ports 21-22)],
 [port-2,
  <class 'pynetbox.models.dcim.Termination'> <> <class 'pynetbox.models.dcim.Termination'>,
  port-2],
 [pair-11 (ports 21-22),
  <class 'pynetbox.models.dcim.Termination'> <> <class 'pynetbox.models.dcim.Termination'>,
  Ethernet11]]

```

So it looks like maybe I was mistaken or confused about only the middle (not the first or last) having the issue, it may be all of the items returned from the trace until they are accessed in some way. You can see here how the info is filled out once the item is accessed directly: 
```
In [2]: trace = iface.trace()                                                                                                     

In [3]: trace                                                                                                                     
Out[3]: 
[[em1,
  <class 'pynetbox.models.dcim.Termination'> <> <class 'pynetbox.models.dcim.Termination'>,
  pair-11 (ports 21-22)],
 [port-2,
  <class 'pynetbox.models.dcim.Termination'> <> <class 'pynetbox.models.dcim.Termination'>,
  port-2],
 [pair-11 (ports 21-22),
  <class 'pynetbox.models.dcim.Termination'> <> <class 'pynetbox.models.dcim.Termination'>,
  Ethernet11]]

In [4]: trace[0][1]                                                                                                               
Out[4]: em1 <> pair-11 (ports 21-22)

In [5]: trace                                                                                                                     
Out[5]: 
[[em1, em1 <> pair-11 (ports 21-22), pair-11 (ports 21-22)],
 [port-2,
  <class 'pynetbox.models.dcim.Termination'> <> <class 'pynetbox.models.dcim.Termination'>,
  port-2],
 [pair-11 (ports 21-22),
  <class 'pynetbox.models.dcim.Termination'> <> <class 'pynetbox.models.dcim.Termination'>,
  Ethernet11]]

```

pynetbox/models/dcim.py
raddessi: I tried, it wants args though :/

```
  File "./test.py", line 33, in <module>
    main()
  File "./test.py", line 26, in main
    trace = iface.trace()
  File "/home/raddessi/dev/spotx/sysops/python-modules/spotx_sysops/src/pynetbox/models/dcim.py", line 57, in trace
    return_obj_class(hop_item_data)
TypeError: __init__() missing 2 required positional arguments: 'api' and 'endpoint'
```

pynetbox/models/dcim.py
raddessi: By moving this to inside the trace function it is muuuuch clearer now.

tests/test_dcim.py
```diff
@@ -325,6 +325,28 @@ def test_get_all(self, mock):
...
         )
 
+    @patch(
+        "pynetbox.core.query.requests.sessions.Session.get",
+        side_effect=[
+            Response(fixture="dcim/interface.json"),
+            Response(fixture="dcim/interface_trace.json"),
+            Response(fixture="dcim/cable.json"),
+            Response(fixture="dcim/cable.json"),
+            Response(fixture="dcim/cable.json"),
```
raddessi, line 335: This is the initial trace, and then the cable information being populated for each of the 3 hops in the trace.

raddessi (commented):
OK I think this is much better.

tests/fixtures/dcim/interface_trace.json
```diff
@@ -0,0 +1,118 @@
...
+        {
+            "id": 5583,
+            "url": "http://localhost:8000/api/dcim/front-ports/5583/",
+            "device": {
+                "id": 4430,
+                "url": "http://localhost:8000/api/dcim/devices/4430/",
+                "name": "test1-patchpanel1",
+                "display_name": "test1-patchpanel1"
+            },
+            "name": "pair-11 (ports 21-22)",
```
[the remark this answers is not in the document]
zachmoody: I don't _think_ it's terribly important those match up across fixtures.

raddessi:
Right on! Is there anything left you would like done here then? If not then I feel like it's ready to go in.

raddessi:
Hold on this actually, I broke functionality in the last update. Will fix asap.

pynetbox/models/dcim.py
```diff
@@ -13,12 +13,67 @@
...
+    def trace(self):
+        req = Request(
+            key=str(self.id) + "/trace" if not self.url else None,
+            base=self.endpoint.url,
+            token=self.api.token,
+            session_key=self.api.session_key,
+            http_session=self.api.http_session,
+        )
+        ret = []
+        # from IPython import embed
```
raddessi, line 35: this needs to be removed as well

raddessi (commented):
another change

pynetbox/models/dcim.py
raddessi: done

raddessi:
OK that should be resolved now. I'm not sure why I had the else statement in there to begin with.

zachmoody:
Nice work! I'd like to put it through its paces as well before merging. This month's pretty crazy for me, but I should be able to get to it before month's end.

raddessi:
Yeah ok no problem. Do you have any plans for an integration testing suite?

zachmoody:
> Yeah ok no problem. Do you have any plans for an integration testing suite?

No serious plans, but that'd be an improvement over the fixtures we're using now. :thinking:

raddessi:
Right on.. how would you feel about me starting a WIP PR to try to set up a basic system using pytest-docker then? If its beyond what you want to deal with right now that's fine it can wait until some point in the future, no big deal.

markkuleinio:
A side note: does the proposed code work if the NetBox server address is like https://appserver.example.com/netbox? The URL paths will then be `/netbox/api/dcim/something/something`.

Checking because we fixed that kind of issue somewhere else in the code recently.

raddessi:
Oh now that is a great point. You're right that would break this. I'll amend asap.

markkuleinio:
@zachmoody An idea (that you maybe already had in the earlier discussion): maybe we should parse and store the "NetBox path" in an `Api` instance attribute or something like that? To prevent the need to parse the same thing in several places.

zachmoody:
> Right on.. how would you feel about me starting a WIP PR to try to set up a basic system using pytest-docker then? If its beyond what you want to deal with right now that's fine it can wait until some point in the future, no big deal.

Yeah, go for it. Was thinking about tackling it myself once I cleared out some of the PR backlog. Will try and help out once you get a PR going, assuming I don't run into any surprises this week.

raddessi:
This is actually much better as I didn't before realize that the uri I needed is already being passed in as `self.endpoint.url`, so this issue should be fixed now and is less confusing as well.

zachmoody:
> This is actually much better as I didn't before realize that the uri I needed is already being passed in as `self.endpoint.url`, so this issue should be fixed now and is less confusing as well.

Nice, been playing around with your PR today. I noticed this change caused the Cable objects in the trace to start coming across as generic Records. Might be worth testing that in some unit tests.

Also been thinking about ways we can avoid requiring additional calls to the cables endpoint since we only get a summarized cable object in the trace response. I'm leaning towards adjusting the Cable object's str method to return something that doesn't require terminations be present. E.g.
``` python
def __str__(self):
    if all(["name" in dict(i) for i in (self.termination_a, self.termination_b)]):
        return "{} <> {}".format(self.termination_a, self.termination_b)
    return "Cable #{}".format(self.id)
```

That's a little more inline with the UI experience, and would be more useful in any other situations (if they exist) where a cable object is sent without terminations. Though if anyone has a better approach, I'm all for it. Just don't like making any more calls to NetBox than are necessary.

raddessi:
> Nice, been playing around with your PR today. I noticed this change caused the Cable objects in the trace to start coming across as generic Records. Might be worth testing that in some unit tests.

Oh wow, ok I did not see that yet. Yeah I'll try to play around with it to reproduce.

> Also been thinking about ways we can avoid requiring additional calls to the cables endpoint since we only get a summarized cable object in the trace response.

That is really nice since as you said it does match with the ui experience. I like it, if we need the full info for a report we can just query the terminations instead of printing the cable. Muuuch cleaner. Also, many fewer calls like you said - it's an all around win. Feel free to push up unless you want me to.

FYI I sent in a PR for integration testing at https://github.com/digitalocean/pynetbox/pull/315 - this change may be something that be integration tested especially with the upcoming cable tracing changes..?

zachmoody:
> Oh wow, ok I did not see that yet. Yeah I'll try to play around with it to reproduce.

Ok, this is fixed up in 4a0d8ca.

> That is really nice since as you said it does match with the ui experience. I like it, if we need the full info for a report we can just query the terminations instead of printing the cable. Muuuch cleaner. Also, many fewer calls like you said - it's an all around win. Feel free to push up unless you want me to.

.trace() output looks like this now.
```
In [28]: x.trace()
Out[28]: [[eth2, Cable #2, fp0], [rp0, Cable #5, rp0], [fp0, Cable #4, eth2]]
```
Calling `x.trace()[1]` or `x.trace()[1].full_details()` will populate the terminations, but that's kind of a round about way of getting data that's already in `x.trace()[0]` and `x.trace()[2]`.

> FYI I sent in a PR for integration testing at #315 - this change may be something that be integration tested especially with the upcoming cable tracing changes..?

Nice! Had a cursory look, seemed off to a good start. Hope to dig into it more later.

zachmoody (approved):

PR #282, opened 2020-08-16
Implement Record.__getitem__()

### LINKED ISSUE #281: `__getitem__()` for Record sends back its argument
Hey!

The code for `__getitem__()` in `response.py` is:

```python
    def __getitem__(self, item):
        return item
```

This is quite confusing if you happen to use the dict notation instead of the attribute notation. I would suggest to either support the dict notation properly (not tested and `__setitem__()` should be implemented too):

```diff
diff --git a/pynetbox/core/response.py b/pynetbox/core/response.py
index 49d9dc57cdeb..f2c8aed4f84a 100644
--- a/pynetbox/core/response.py
+++ b/pynetbox/core/response.py
@@ -186,14 +186,11 @@ class Record(object):
         In order to prevent non-explicit behavior,`k='keys'` is
         excluded because casting to dict() calls this attr.
         """
-        if self.url:
-            if self.has_details is False and k != "keys":
-                if self.full_details():
-                    ret = getattr(self, k, None)
-                    if ret or hasattr(self, k):
-                        return ret
-
-        raise AttributeError('object has no attribute "{}"'.format(k))
+        if k != "keys":
+            try:
+                return self[k]
+            except KeyError:
+                raise AttributeError('object has no attribute "{}"'.format(k))
 
     def __iter__(self):
         for i in dict(self._init_cache):
@@ -207,8 +204,15 @@ class Record(object):
             else:
                 yield i, cur_attr
 
-    def __getitem__(self, item):
-        return item
+    def __getitem__(self, k):
+        if self.url:
+            if self.has_details is False:
+                if self.full_details():
+                    ret = getattr(self, k, None)
+                    if ret:
+                        return ret
+
+        raise KeyError('object has not key "{}"'.format(k))
 
     def __str__(self):
         return getattr(self, "name", None) or getattr(self, "label", None) or ""
```

Or to remove this `__getitem()__` to not let the user retrieve obviously incorrect results.

I was being started on the first solution, but maybe I am missing something?

vincentbernat:
The current implementation returns the provided key. This is confusing
for the user as it is unexpected. Instead, we turn ourselves into a
dict and return the expected value. This is not really efficient but
the whole logic is already implemented in other functions.

Fix #281.

### DIFF
--- a/pynetbox/core/response.py
+++ b/pynetbox/core/response.py
@@ -207,8 +207,8 @@ class Record(object):
             else:
                 yield i, cur_attr
 
-    def __getitem__(self, item):
-        return item
+    def __getitem__(self, k):
+        return dict(self)[k]
 
     def __str__(self):
         return getattr(self, "name", None) or getattr(self, "label", None) or ""
--- a/tests/unit/test_response.py
+++ b/tests/unit/test_response.py
@@ -11,6 +11,36 @@ else:
 
 
 class RecordTestCase(unittest.TestCase):
+    def test_attribute_access(self):
+        test_values = {
+            "id": 123,
+            "units": 12,
+            "nested_dict": {"id": 222, "name": "bar"},
+            "int_list": [123, 321, 231],
+        }
+        test_obj = Record(test_values, None, None)
+        self.assertEqual(test_obj.id, 123)
+        self.assertEqual(test_obj.units, 12)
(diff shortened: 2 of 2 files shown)


zachmoody:
Well that's elegant. Nice! Thanks for the PR.

MattIPv4:
Hey, @vincentbernat - Thanks a ton for this sweet PR! 😄

Issue #224, opened 2020-02-20
getting nonexistent device (or any other entity) by id causing exception


Dimaqa:
Hello!
`nb.dcim.devices.get(26) `
throws
`pynetbox.core.query.RequestError: The requested url: /api/dcim/devices/26/ could not be found.`
But /api/dcim/devices/?id=26 returns None. 
And i cant call 
`nb.dcim.devices.get(id=26)`
 because it throws  
`ValueError: A reserved ('id', 'pk', 'limit', 'offset') kwarg was passed. Please remove it try again.`

zachmoody:
you'll want to use the former syntax, and `/api/dcim/devices/26/` looks like a properly formed url. Does it pull up for you in the browser?

Dimaqa:
id=26 is just an example of not existing device.
My idea was, that get() function should not raise exceptions, so you can use it like
```
if (device = nb.dcim.devices.get(id=26)):
   device.delete()
```
I think the easiest way to solve this just to let use RESERVED_KWARGS if no args given:
```
if any(i in RESERVED_KWARGS for i in kwargs) and not agrs:
            raise ValueError(
                "A reserved {} kwarg was passed. Please remove it "
                "try again.".format(RESERVED_KWARGS)
            )
```
Or at least change request error on something like 'no device with such id'

zachmoody:
Istr `id` not being a valid filter in previous NetBox version, but I guess that's changed (or I didn't test well enough to begin with 😄 ). Assuming id wasn't in RESERVED_KWARGS though, `.get(id=26)` and `.get(26)` are practically the same. What's the motivation for needing to pass the id kwarg?

Dimaqa:
So get request will return None and wont raise error, because query will be `/api/dcim/devices/?id=26 `not `/api/dcim/devices/26/`. Or it will be` /api/dcim/devices/26/` anyway?

dsg-anthony:
This also happens when you use filter
Example: getting a list of VM ids from the ipam_services endpoint and only wanting VMs from cluster id 4.

`virtualization.virtual_machines.filter(id=vm_id, cluster_id=4)`

Dimaqa:
So i've found a funny solution.
If you dont want your get or filter to cause the exception you can use query directly.
For example, if device with id = 26 does not exist
`nb.dcim.devices.get(26)`
will raise error
`pynetbox.core.query.RequestError: The requested url: /api/dcim/devices/100/ could not be found.`
but 
[...]

PR #292, opened 2020-09-25
Fix of #224

Dimaqa:
So #224 still happens, becaise exception happens in req.get(), not in Request creation
This will fix it


Dimaqa:
But after such fix endpoint class will also return None in bad requests with typos, such as dcim.deivces and etc :(

zachmoody (approved):
ah, right, thanks for catching that.

PR #303, opened 2020-12-08
Do not create new endpoint if plugin returns URL.

### LINKED ISSUE #302: Wrong URL with Response on plugins Endpoint.
When using pynetbox with plugins, only creating a listing works. When I want for example deletion, pynetbox uses the wrong URL.

The underlying issue lies in having `url` set in GET serializer. If it's there, `Record.__init__` does not return endpoint directly but instead tries to return endpoint from gettattr(App, name), which returns new endpoint with wrong URL.
```
        self.endpoint = (
            self._endpoint_from_url(values["url"])
            if values and "url" in values
            else endpoint
        )
```
```
    def _endpoint_from_url(self, url):
        app, name = urlsplit(url).path.split("/")[2:4]
        return getattr(pynetbox.core.app.App(self.api, app), name)
```
This whole new Endpoint does not expect it to be plugin endpoint so it fails to get the correct URL. Instead of `http://127.0.0.1/api/plugins/ip_health_check/check/7` it tries to use `http://127.0.0.1/api/plugins/ip-health-check/7/`.

There is a nasty workaround though, manually changing internal endpoint.base_url:
```     
        response = nb.plugins.ip_health_check.check.filter(ip=ip.id)[0]
        response.endpoint.url = response.endpoint.base_url + "/plugins/ip_health_check/check"
        response.update({"result": True})
        response.delete()
```

The fix though seems really easy once I know all this, just avoid getting new endpoint when working with plugins. Will submit PR shortly.

Thetacz:
Fixes #302 

Tests passed, but looking through the test cases they just test getting installed plugins, not using the plugins API. I do not have time to write new tests to cover this whole case, sorry.

### DIFF
--- a/pynetbox/core/response.py
+++ b/pynetbox/core/response.py
@@ -171,7 +171,7 @@ class Record(object):
         self.default_ret = Record
         self.endpoint = (
             self._endpoint_from_url(values["url"])
-            if values and "url" in values
+            if values and "url" in values and "/api/plugins/" not in values["url"]
             else endpoint
         )
         if values:


zachmoody:
Thanks for the PR, we fixed this in #324.

PR #305, opened 2020-12-08
Added update() capability to endpoint APIs

rstor-d1ng0:
Allows updating existing endpoint objects in NetBox using the PATCH method.


zachmoody:
I think you can accomplish the same thing with the existing API with something like:
```
for i in nb_api.app.endpoint.filter(...):
  i.update({"field": "value"})
```
Let me know if there's a use-case I'm missing though.

PR #307, opened 2020-12-14
fix: tag management with netbox >= 2.9

jbe-dw:
Hello,

This PR is made with the purpose of keeping backward compatibility with previous versions. There's in my opinion another way arround I though about: we can make tags membership of `LIST_AS_SET` conditional to the API version and keep it the way it was.

This would force end users to first create or retrieve tags objects in or from Netbox API and use the records in the append() function. In this case  #291 would just need a version check conditional if i'm correct. But it wouldn't work when user provides a dictionnary.

### DIFF
--- a/pynetbox/core/api.py
+++ b/pynetbox/core/api.py
@@ -101,6 +101,7 @@ class Api(object):
         self.extras = App(self, "extras")
         self.virtualization = App(self, "virtualization")
         self.plugins = PluginsApp(self)
+        self.api_version = None
 
     @property
     def version(self):
--- a/pynetbox/core/endpoint.py
+++ b/pynetbox/core/endpoint.py
@@ -98,7 +98,10 @@ class Endpoint(object):
             threading=self.api.threading,
         )
 
-        return response_loader(req.get(), self.return_obj, self)
+        get_req = req.get()
+        self.api.api_version = req.api_version
+
+        return response_loader(get_req, self.return_obj, self)
 
     def get(self, *args, **kwargs):
         r"""Queries the DetailsView of a given endpoint.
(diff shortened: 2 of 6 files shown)


zachmoody:
Hey, thanks for the PR. I'm a little hesitant to take this approach because (I think) you'll end up beating up the endpoint we're getting the version from if you have a lot of changes. We could cache that value, but I'd like to explore just inspecting the elements of `tags` and handling them based on their type. I hope to have some time next week to dig into that solution, if folks can be patient. If that doesn't pan out, let's cache the version and merge this one.

markkuleinio:
I support the idea of inspecting the tag object type and acting accordingly.

Also, instead of caching the NetBox API version (if needed frequently for any reason, internal to `pynetbox` or for the application reasons) in the first call, I would like to investigate an option to keep an updated API version attribute in the `pynetbox.Api` instance so that from every NetBox response header we would update (if needed) the API version attribute so that we always have an updated NetBox API version at hand. That way we would not introduce a dependency to restart each and every (possibly remote) application component whenever we update the NetBox server, which would be needed if we cache the API version unconditionally. Restarting apps is not a problem in small deployments or when a script runs separate runs, but for example I have various webhook handlers and DHCP automation modules that are running all the time on different platforms, and keeping the running apps synchronized with various changes is getting harder. Time-based caches are not significantly better.

zachmoody:
:100: , that would be nice.

jbe-dw:
Hello

I didn't update the pytest yet but I made some changes to catch the api version from headers after each calls. The changes don't look elegant to me and may probably have a more pythonic version in your mind.

Tell me about the changes, I didn't spend a lot of time for this, I'm still getting confident that you'll soon find some time to do the fix. If you wish an improvement (and make CI green again) to release this fix because you lack of time let me know.

Regards.

markkuleinio:
See also #311

zachmoody:
Thanks again for the PR, but ended up opting for the approach in #311.

PR #308, opened 2020-12-15
Fixes #304: Change Record._endpoint_from_url() to count URL fields from the end

### LINKED ISSUE #304: Unable to use .save() on tenant
Netbox 2.9.10
pynetbox 5.1.0

Creating a tenant works:

tenant = nb.tenancy.tenants.create({'name': 'Tenant', 'slug': 'tenant'})

Trying to set the group on tenant:

tenant.group = 1

Doing tenant.save() results in:

pynetbox.core.query.RequestError: The requested url: https://server1.int.net/ipam-stage/api/api/tenancy/257/ could not be found

doing pprint(dict(tenant)):

{'comments': '',
 'created': '2020-12-08',
 'custom_fields': {},
 'description': '',
 'group': 1,
 'id': 258,
 'last_updated': '2020-12-08T19:28:12.002493Z',
 'name': 'Tenant',
 'slug': 'tenant',
 'tags': [],
 'url': 'https://server1.int.net/ipam-stage/api/tenancy/tenants/258/'}

It sees that pynetbox looses the "tenants" in the url between tenancy and id of tenant

markkuleinio:
This is to support NetBox API base URLs with one or more directories. In those cases the previous code picked incorrect parts from the URL.

Fixes #304.

### DIFF
--- a/pynetbox/core/response.py
+++ b/pynetbox/core/response.py
@@ -280,7 +280,11 @@ class Record(object):
             setattr(self, k, v)
 
     def _endpoint_from_url(self, url):
-        app, name = urlsplit(url).path.split("/")[-4:-2]
+        # Remove the base URL from the beginning
+        url = url[len(self.api.base_url):]
+        if url.startswith("/"):
+            url = url[1:]
+        app, name = urlsplit(url).path.split("/")[:2]
         return getattr(pynetbox.core.app.App(self.api, app), name)
 
     def full_details(self):
--- a/pynetbox/core/response.py
+++ b/pynetbox/core/response.py
@@ -280,7 +280,7 @@ class Record(object):
             setattr(self, k, v)
 
     def _endpoint_from_url(self, url):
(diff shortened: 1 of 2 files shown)


zachmoody:
Ha, was just about to respond to you in the issue saying I hoped to make some time next week to work on the laundry list of issues that have piled up. You beat me to this one :)

Only concern with this approach is the paths that don't conform to what we're expecting like `/ipam/prefixes/:id/available-ips/`. I'm just spit-balling here, but wonder if we anchored the index on `/api/` if we could get more predictable results. :thinking: 

Adding a test for that and the directory URLs in `test_endpoint_from_url` would be great too.

markkuleinio:
Would it be safe to first remove `self.api.base_url` from `url` and then get the app and endpoint after that?

zachmoody (approved):
Good stuff, thanks as always @markkuleinio.

PR #310, opened 2020-12-17
Change Record._endpoint_from_url() to handle stripped http/https default ports

### LINKED ISSUE #309: API .save() calls Fail when API host is in the format host:port
When instantiating pynetbox API with a URL in the format host:port, as below, base_url gets set to http://host:port, but objects returned from the API do not include the :port designation. This is causing truncation of endpoint URLs when calling _endpoint_from_url in response.py

```
nb = pynetbox.api(
    host='localhost:80',
    'token'=mytoken
)
```

`device = nb.dcim.device.get(1)`

device.url will be "http://localhost/api/dcim/devices/1/"
base_url will be "http://localhost:80/api"

Calling `device.save()`

endpoint URL becomes:
im/device instead of dcim/device

markkuleinio:
Fixes #309 (that was caused by the parsing change in #308)

NetBox removes the default http/https (80 or 443 respectively) port from the response object URL even though it is given for any reason in the `pynetbox.Api` instantiation. That makes `Record._endpoint_from_url()` generate incorrect `App` names (it removes too many characters from the given URL) in those cases.

This PR makes `Record._endpoint_from_url()` to remove the default ports so that the `App` name is correctly parsed. Parsing does not look very elegant but works 😐

### DIFF
--- a/pynetbox/core/response.py
+++ b/pynetbox/core/response.py
@@ -280,11 +280,20 @@ class Record(object):
             setattr(self, k, v)
 
     def _endpoint_from_url(self, url):
+        api_base_url = self.api.base_url
+        if not url.startswith(api_base_url):
+            # Let's account for stripped http..:80 or https..:443
+            b = urlsplit(api_base_url)
+            if (b.scheme == "http" and b.netloc.endswith(":80")) or (
+                b.scheme == "https" and b.netloc.endswith(":443")
+            ):
+                # Rewrite base_url without the default port
+                api_base_url = "{}://{}{}".format(
+                    b.scheme, b.netloc.split(":")[0], b.path,
+                )
         # Remove the base URL from the beginning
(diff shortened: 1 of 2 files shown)


markkuleinio:
Or, should we just remove the default ports from the URL right in the `pynetbox.Api` instantiation and forget all this mangling in `Record._endpoint_from_url()`?

zachmoody:
> Or, should we just remove the default ports from the URL right in the `pynetbox.Api` instantiation and forget all this mangling in `Record._endpoint_from_url()`?

Yeah, if we're going to take on sanitizing input, we should probably do it there. I can whip up a PR for it a bit later.

markkuleinio:
Ok, I'll close this (https://tools.ietf.org/html/rfc3986#section-3.2.3 was mentioned in #309).

PR #312, opened 2020-12-18
Cleaning: remove url_param_builder() and unused requests import

markkuleinio:
Cleaning: remove `url_param_builder()` and unused `requests` import.

`tests/test_query.py` only contained test for the deleted `url_param_builder()` so the whole file was deleted in the process.


zachmoody:
:tada: Clean up! Thanks!

Issue #322, opened 2020-12-30
API .update() fails when API host is in the format "http://ipaddress". Endpoint API URL has missing characters


thecompletenoob:
In reference to Issue #309 .update() still fails for me when I instantiate the api with "http://ipaddress".
 @markkuleinio suggested I raise a new issue with more information.
The only thing that works for me is "http://localhost" or "http://127.0.0.1" which I can't explain why.

My setup is:
Ubuntu 18.04.5 LTS (GNU/Linux 4.15.0-128-generic x86_64)
nginx version: nginx/1.14.0 (Ubuntu)
Netbox v2.10.2
pynetbox==5.2.1
requests==2.25.1
six==1.11.0

This is the sample code I used:
>		import pynetbox
>		
>		netboxUrl = 'http://10.92.105.252'
>		token = 'mytoken'
>		
>		nb = pynetbox.api(url=netboxUrl, token=token)
>		
>		ipNew = nb.ipam.ip_addresses.create({"address": "192.168.7.1/24"})
>		
>		tenantNew = nb.tenancy.tenants.create({"name":"Test", "slug":"test"})
>		
>		ip = nb.ipam.ip_addresses.get(q="192.168.7.1/24")
>		
>		tenant = nb.tenancy.tenants.get(name="Test")
>		
>		try:
>		    ip.update({"tenant":tenant.id})
>		except pynetbox.RequestError as e:
>		    print(f"\nCould not update the ip, error: {str(e)}\n")
>

This is the error with the missing characters in the URL:
>		Could not update the ip, error: The requested url: http://10.92.105.252/api/m/ip-addresses/922/ could not be found.

Similar error for similar code but with the dcim/sites api:
>		Could not update the site, error: The requested url: http://10.92.105.252/api/m/sites/16/ could not be found.

I also realized that when i used a 192.168.1.5 during instantiation, the number of missing characters had reduced.
>		Could not update the ip, error: The requested url: http://192.168.1.5/api/pam/ip-addresses/189/ could not be found.

Kindly lemme know if more information is needed. Thanks for the support.

markkuleinio:
What does `ip.url` show?

What is your nginx configuration?

What is your Python version?

thecompletenoob:
@markkuleinio 
ip.url
>             ip.url: http://localhost/api/ipam/ip-addresses/923/

Python 3.6.9

nginx configuration:
>		server {
>		    listen 80;
>		
>		    server_name localhost;
>		
[...]

PR #323, opened 2021-01-02
Change Record._endpoint_from_url() path parsing to ignore host:port

markkuleinio:
This PR changes the logic in `Record._endpoint_from_url()`: with this change it does not matter what's in the host:port part of the URL as we always check the path of `Api.base_url`, and if it contains some extra leading directories, those extra directories will be removed from `url` before parsing the `app` and `name`.

There are two sides in this:
1. This enables the path parsing to work even if the host:port is for some reason different in the NetBox-returned URLs and in the `pynetbox` user-supplied `Api.base_url`.
2. This also masks out some reverse proxy configuration errors like in #322 (but that was the original behaviour before #308 introduced in `pynetbox` 5.1.1).

### DIFF
--- a/pynetbox/core/response.py
+++ b/pynetbox/core/response.py
@@ -280,11 +280,13 @@ class Record(object):
             setattr(self, k, v)
 
     def _endpoint_from_url(self, url):
-        # Remove the base URL from the beginning
-        url = url[len(self.api.base_url) :]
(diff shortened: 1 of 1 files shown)


zachmoody (approved):
:+1: Thanks! Assuming we can skip normalization now.

PR #325, opened 2021-01-26
Fix for interface tracing

raddessi:
I missed a mistake when this was merged, the url split changes we made had some bad assumptions in them, like it did not take in to account the possible presence of a port number in `self.api.base_url` that would throw off indexing. This would result in everything from a `.trace()` response being a generic `Record` since we were looking incorrectly for keys like `interfaces/1234` in the `uri_to_obj_class_map` map.

I also added front and rear port tracing which I think was added since I had tested.


raddessi:
Linter failures should be addressed by https://github.com/digitalocean/pynetbox/pull/326

raddessi:
I need to continue this, I was able to reproduce some failures sporadically locally but I haven't had time to continue yet.

raddessi:
Looks like the tests are failing because of the issues I saw I caused by not pinning the release tags properly.. at least I think that is the cause. https://github.com/digitalocean/pynetbox/pull/327 should fix that I believe

PR #327, opened 2021-01-29
Fixed failing tests on master

raddessi:
Since they just released 1.0 I noticed I didn't pin the repo version properly, sorry about that. This in it's current state should fix the existing tests. If I have time before you get to this I'll try to actually update the version to 1.0 as well.

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -496,9 +496,18 @@ def netbox_service(
     netbox_integration_version = request.param
 
     # `port_for` takes a container port and returns the corresponding host port
-    port = docker_services.port_for(
-        "netbox_v%s_nginx" % str(netbox_integration_version).replace(".", "_"), 8080
+    docker_service_name = "netbox_v%s_nginx" % str(netbox_integration_version).replace(
+        ".", "_"
     )
+    try:
+        port = docker_services.port_for(docker_service_name, 8080)
+    except Exception:
+        docker_ps_stdout = subp.check_output(["docker", "ps", "-a"]).decode("utf-8")
+        raise KeyError(
+            "Unable to find a docker service matching the name %s. Running containers:"
+            " %s" % (docker_service_name, docker_ps_stdout)
+        )
+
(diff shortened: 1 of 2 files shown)


tests/integration/conftest.py
```diff
@@ -133,6 +133,12 @@ def netbox_docker_repo_dirpaths(pytestconfig, git_toplevel):
             subp.check_call(
                 ["git", "fetch"], cwd=repo_fpath, stdout=subp.PIPE, stderr=subp.PIPE
             )
+            subp.check_call(
+                ["git", "reset", "--hard"],
+                cwd=repo_fpath,
+                stdout=subp.PIPE,
+                stderr=subp.PIPE,
+            )
```
raddessi, line 141: This was needed to fix some changes that were made when I ran black over everything.. it's probably just a good idea in general though.

raddessi:
Hmm.. maybe this is due to memory limits. I can't get this to fail locally any more. Let me bump down the versions we test concurrently..

raddessi:
No change there.. This is really odd. Tests continue to run fine locally.

raddessi:
And on https://github.com/digitalocean/pynetbox/pull/327/commits/ce0c31d6ecfec24318f6fffaf0c05b5f6b78ad05 the tests only failed on 3.8.. this seems real flaky whatever it is.

raddessi:
Woot! I have _no_ clue exactly what broke since it was building fine off these same tags before.. but updating netbox 2.10 to the netbox-docker tag 1.0.1 seems to have fixed that now. Sweeeeeeeet.

raddessi:
Ok.. i think i know why this failed. Docker looks for the right image from a tag at time of pull but then caches that version (right? I think?) so my local version was something like 2.10.1 and passing tests but github was building against 2.10.4 and failing. Maybe there is a way to force it to check for updates.. maybe i should make ot.more clear what patch version is being tested also.

zachmoody:
:+1: Thanks!

PR #343, opened 2021-04-02
Implement more integration tests

zachmoody:
Takes a more-or-less standardized approach to integration testing. Only covering most of the dcim and ipam apps with this PR but looking to expand that in the future. Also implements a way to use an existing instance of NetBox by way of the `--url-override` arg to hopefully speed up dev iterations that may use the integration tests. Since taking this approach I don't think we need the Faker or Device Template dependencies so I pulled them.

cc: @raddessi


raddessi:
Right on! This looks good

raddessi:
I'll have to play around with it but I think this is very extendable. Do you feel comfortable with people adding tests with this structure then now or are you still feeling it out?

zachmoody:
I think I'm comfortable extending this approach. One thing I do want to try tackling at some point is a few tests on `.all()` and `.filter()` with threading enabled. I think there's at least a couple ways to approach that though.

PR #358, opened 2021-04-06
Add a __next__ method to complete the iterator protocol

### LINKED ISSUE #354: Iterator protocol not fully implemented, missing __next__
```
❯ ipython3
Python 3.9.2 (default, Feb 28 2021, 17:03:44) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.20.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import pynetbox, subprocess, os

In [2]: api = pynetbox.api('https://netbox.xy.zz', token=os.environ['NETBOX_TOKEN'])

In [3]: devices = api.dcim.devices.filter(name='doesnotexist')

In [4]: next(devices)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-4-e3dedbd77ad5> in <module>
----> 1 next(devices)

TypeError: 'RecordSet' object is not an iterator
```

This helper depends on a `__next__` method on the generator object.

jqueuniet:
Implement the `__next__` method to complete the iterator protocol

Fix #354

### DIFF
--- a/pynetbox/core/endpoint.py
+++ b/pynetbox/core/endpoint.py
@@ -133,7 +133,7 @@ class Endpoint(object):
             key = None
 
         if not key:
-            return next(iter(self.filter(**kwargs)), None)
+            return next(self.filter(**kwargs), None)
 
         req = Request(
             key=key,
@@ -143,7 +143,7 @@ class Endpoint(object):
             http_session=self.api.http_session,
         )
         try:
-            return next(iter(RecordSet(self, req)), None)
+            return next(RecordSet(self, req), None)
         except RequestError as e:
             if e.req.status_code == 404:
                 return None
--- a/pynetbox/core/response.py
+++ b/pynetbox/core/response.py
@@ -78,16 +78,16 @@ class RecordSet(object):
         self._response_cache = []
 
     def __iter__(self):
-        for i in self._response_cache:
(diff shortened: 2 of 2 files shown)


zachmoody:
I struggled a bit on use-cases for making it an iterator, so I left it out, but I can see some use for it now. I added a couple tweaks, let me know what you think. I was getting a bug when I called `len()` on it before stepping through with `next()` that was related to the cache. I think these commits should fix it.

jqueuniet:
I'll admit it's pretty rare, but I have a few use-cases where I don't want to iterate over the whole set, just test the record length and extract the first one or something like that. The one for which I had to implement this change is checking a device interface at boot time for an iPXE script, where I query by MAC, I should only get one result, but Netbox can store duplicated interface MAC as it is valid in some cases. So, I don't want to crash my call with a `get()` on devices like that, just to log a warning as this device is probably badly inventoried.

I don't have an issue with your changes, they're probably better suited to your implementation. Only wanted to start the ball rolling so to speak.

zachmoody:
Ah, understood. You might've seen in the diff, `.get()`'s behavior changed as a result of this too. It won't throw an exception anymore if there's more than one item in a response now. Might be a good fit again for your use-case.

PR #388, opened 2021-06-24
Fix object output for `users` in permissions

markkuleinio:
The problem:

```
>>> permission = list(netbox.users.permissions.all())[0]
>>> from pprint import pprint as ppr
>>> ppr(dict(permission))
{'actions': ['view'],
 'constraints': None,
 'description': '',
 'display': 'ViewSites',
 'enabled': True,
 'groups': [],
 'id': 1,
 'name': 'ViewSites',
 'object_types': ['dcim.site'],
 'url': 'http://netbox-test.lein.io/api/users/permissions/1/',
 'users': [{'display': 'test1',
            'id': 2,
            'url': 'http://netbox-test.lein.io/api/users/users/2/',
            'username': 'test1'}]}
>>> permission.users
[]
>>>
```
Meaning that the username "test1" is shown as an empty string. That is because `Record.__str__()` only tries to get fields `name` and `label`, and returns empty string if neither of those is found.

This PR addresses this issue by modifying `Record.__str__()` to also try `username` field when returning the string representation of an object.

After the change:

```
>>> permission.users
[test1]
>>>
```

### DIFF
--- a/pynetbox/core/response.py
+++ b/pynetbox/core/response.py
@@ -272,7 +272,7 @@ class Record(object):
         return (
             getattr(self, "name", None)
             or getattr(self, "label", None)
-            or getattr(self, "username", None)
+            or getattr(self, "display", None)
             or ""
         )
 
--- a/pynetbox/core/response.py
+++ b/pynetbox/core/response.py
@@ -269,7 +269,12 @@ class Record(object):
         return dict(self)[k]
 
     def __str__(self):
-        return getattr(self, "name", None) or getattr(self, "label", None) or ""
+        return (
+            getattr(self, "name", None)
+            or getattr(self, "label", None)
+            or getattr(self, "username", None)
+            or ""
+        )
 
     def __repr__(self):
         return str(self)
--- a/tests/fixtures/users/permission.json
+++ b/tests/fixtures/users/permission.json
@@ -1,4 +1,9 @@
 {
     "id": 1,
-    "name": "permission1"
+    "name": "permission1",
+    "users": [
+        {
+            "username": "user1"
+        }
+    ]
 }
--- a/tests/fixtures/users/permission.json
+++ b/tests/fixtures/users/permission.json
@@ -3,6 +3,7 @@
     "name": "permission1",
     "users": [
         {
+            "display": "user1",
             "username": "user1"
         }
     ]
--- a/tests/test_users.py
+++ b/tests/test_users.py
@@ -95,3 +95,13 @@ class GroupsTestCase(Generic.Tests):
 
 class PermissionsTestCase(Generic.Tests):
     name = "permissions"
+
+    @patch(
+        "requests.sessions.Session.get",
+        return_value=Response(fixture="users/permission.json"),
+    )
+    def test_username(self, _):
+        permission = nb.permissions.get(1)
+        self.assertEqual(len(permission.users), 1)
+        user = permission.users[0]
+        self.assertEqual(str(user), "user1")


zachmoody:
Is it possible to go for a custom model here instead of changing the default behavior?

markkuleinio:
Oh, that sounds like a good idea, but I have currently no idea how that works 🤔😂

There is a `Users(Record)` class in `models/users.py`, how should we use it in this case?

zachmoody:
Ah ok, it's already there and doing the right thing. 

```python
class Users(Record):
    def __str__(self):
        return self.username
```

Looks like we need a custom model for permissions letting it know that field is a list of User objects. It being a list might present some problems though, not sure we've had to do that before :thinking:

markkuleinio:
What's the meaning of this in `Record._parse_values()`:

https://github.com/netbox-community/pynetbox/blob/4efcab008ad34d756c06a8329cec7b4c0de8aaa2/pynetbox/core/response.py#L314-L315

Does it sometimes return something else than `None`? In my small test `lookup` was just `None` all the time.

Anyway, would a lookup table be enough for a solution?

```
from pynetbox.models.users import Users # and other relevant custom models
_field_model_lookup = {
    "users": Users,
}
...
if isinstance(v, dict):
    field_model = _field_model_lookup.get(k, Record)
# and the same in list_parser()
```

... except that is does not work due to circular imports: `response.py` would be importing `users.py`, which in turn imports `response.py`...

I guess the current model structure is something that is based on Django ideology so I don't know what's the right+clean solution off the bat. Collecting all the models in the same source files and rewriting the model lookup logic in `app.py` and `endpoint.py`? Or moving `Record` in a separate file and moving the `_parse_values()` logic in a standalone function in `response.py`?

I'll create an issue for this so that we can link also the other possible PRs to that.

markkuleinio:
Created issue #389 for this

zachmoody:
> Does it sometimes return something else than `None`? In my small test `lookup` was just `None` all the time.

Yeah, if memory serves, it shouldn't be none when a model has an attribute with an explicit reference to another model. E.g. https://github.com/netbox-community/pynetbox/blob/ace10614744480b8f8ce7921374936613d17ea68/pynetbox/models/dcim.py#L88-L91

Like if this wasn't a list, I think we could build out a Permissions class under Users and bring it in something like this:

```python
class Permissions(Record):
    users = Users
...
```
Haven't had a chance to think about it too much, but I think you're on the right track in #390 with fixing up list_parser. It might be as simple as just doing a similar lookup there.

markkuleinio:
@zachmoody Do you have have a good idea how to tackle this forward? New data models are being introduced and they have the same issue:

```
In [14]: list(netbox31beta.ipam.asns.all())
Out[14]: []

In [15]: for asn in netbox31beta.ipam.asns.all():
    ...:     pprint(dict(asn))
    ...:
{'asn': 64800,
 'created': '2021-11-21',
 'custom_fields': {},
 'description': '',
 'display': 'AS64800',
 'id': 1,
 'last_updated': '2021-11-21T11:36:46.344524+02:00',
 'rir': 1,
 'site_count': 0,
 'tags': [],
 'tenant': None,
 'url': 'http://netbox31beta.lein.io/api/ipam/asns/1/'}
```

Not a major problem in my production applications though, only in the development stage when executing the commands interactively, still causes some raised eyebrowns 😃

markkuleinio:
@zachmoody Please see PR #390, I've attempted to add support for custom model lists:

```
class Permissions(Record):
    users = [Users]
```

I will now modify _this_ PR to use the `display` field as a last resort to support any future NetBox models that don't yet have a custom model in `pynetbox`.

markkuleinio:
One more comment on this PR. `test_users.py` might need some improvements now, so maybe it's best to have #390 merged in first (if accepted), and I could then iron out the tests in this PR or in a new PR if better suited.

zachmoody:
Hey @markkuleinio, apologies, haven't been able to keep up with the project as of late. Should have time to review all these next week though.

markkuleinio:
Replaced by #419

PR #390, opened 2021-06-26
Add custom model name lookup to fix #389

### LINKED ISSUE #389: `users` in permissions is represented as empty strings
The problem:

```
>>> permission = list(netbox.users.permissions.all())[0]
>>> from pprint import pprint as ppr
>>> ppr(dict(permission))
{'actions': ['view'],
 'constraints': None,
 'description': '',
 'display': 'ViewSites',
 'enabled': True,
 'groups': [],
 'id': 1,
 'name': 'ViewSites',
 'object_types': ['dcim.site'],
 'url': 'http://netbox-test.lein.io/api/users/permissions/1/',
 'users': [{'display': 'test1',
            'id': 2,
            'url': 'http://netbox-test.lein.io/api/users/users/2/',
            'username': 'test1'}]}
>>> permission.users
[]
>>>
```
Meaning that the username "test1" is shown as an empty string. That is because `Record.__str__()` only tries to get fields `name` and `label`, and returns empty string if neither of those is found.

markkuleinio:
Fixes #389 

The `list_parser()` function in `Record._parse_values()` is modified to support lists of custom models, defined like this:

```
class Permissions(Record):
    users = [Users]
```

Running example:

```
>>> perm = list(nb.users.permissions.all())[0]
>>> pprint(dict(perm))
{'actions': ['view'],
 'constraints': {'status': 'test'},
 'description': '',
 'display': 'TestPermission',
 'enabled': False,
 'groups': [],
 'id': 1,
 'name': 'TestPermission',
 'object_types': ['circuits.circuit'],
 'url': 'http://netbox31beta.lein.io/api/users/permissions/1/',
 'users': [{'display': 'admin',
            'id': 1,
            'url': 'http://netbox31beta.lein.io/api/users/users/1/',
            'username': 'admin'}]}
>>> perm.users
[admin]
>>> type(perm.users[0])
<class 'pynetbox.models.users.Users'>
>>>
```

---

The original PR text left here for the memory of all the unsuccessful attempts: 😁

The model name lookup table is populated in the model source file (`users.py`) to prevent circular imports (we cannot import `Users` in `response.py` because `users.py` already imports `response.py`).

Works now:

```
>>> permission = list(netbox.users.permissions.all())[0]
>>> permission.users
[test1]
>>>
```


Dimaqa:
I believe the idea was to pass return_obj to Endpoint class and implement full Permission model not only users field.
So after you define 
```
class User:
     def __str__(self):
        return self.username

class Permission:
     users = User
```
and somehow pass Permission model to permissions Endpoint you get the same behaviour as with default defined models.
This implementation would also be really helpful with plugins integration.

Dimaqa:
I think you just need to modify `_lookup_ret_obj` in `Endpoint` class

Dimaqa:
I think you just need to find a way to pass your new model to `App` class or modify `_lookup_ret_obj` in `Endpoint` class

markkuleinio:
Any new comments anyone? @Dimaqa @zachmoody 

Is it a problem that `Permission.users` is a list of `User` objects, not just one `User`?

Lost track of all these classes and models already 😄 but thinking that something needs to be figured out to support modern NetBox versions. Is it then dealing with all kinds of custom models, or fixing the `__str__()` output with the `display` field as a last resort in https://github.com/netbox-community/pynetbox/blob/33a7fb98d5b5671e8c5a555b859290d9b47b2681/pynetbox/core/response.py#L316 (because that's the common and good enough (for me, I think now, but untested) solution for NetBox 2.11+ objects).

@Dimaqa feel free to open your own PR if you have a better fix in mind.

markkuleinio:
Not sure how @zachmoody gets the notifications for a new commit in an existing PR so here is a ping for you 😁 (see the modified message on top of this PR)

zachmoody (approved):

PR #406, opened 2021-09-20
Partially fixes #316 - adds bulk update to endpoints

### LINKED ISSUE #316: Bulk delete and update support
In netbox 2.10 REST API Bulk deletion and updation were released.
So i see 2 solutions to implemetnt this:
1. Lazy one: define .delete() .update() functions on Endpoint, which will recieve array of dicts.
2. Implement queryset class similar to django and make response_loader function return it, istead of simple array.
I think thats not really hard either way. 

Whats your thoughts on this?

steffenschumacher:
Kid slept through and it wasn't all that much..

### DIFF
--- a/pynetbox/core/endpoint.py
+++ b/pynetbox/core/endpoint.py
@@ -312,6 +312,65 @@ class Endpoint(object):
             return [self.return_obj(i, self.api, self) for i in req]
         return self.return_obj(req, self.api, self)
 
+    def update(self, objects):
+        r"""Bulk updates existing objects on an endpoint.
+
+        Allows for bulk updating of existing objects on an endpoint.
+        Objects is a list whic contain either json/dicts or Record
+        derived objects, which contain the updates to apply.
+        If json/dicts are used, then the id of the object *must* be
+        included
+
+        :arg list objects: A list of dicts or Record.
+
+        :returns: True if the update succeeded
+
+        :Examples:
+
+        Updating objects on the `devices` endpoint:
+
+        >>> device = netbox.dcim.devices.update([
+        ...    {'id': 1, 'name': 'test'},
+        ...    {'id': 2, 'name': 'test2'},
+        ... ])
+        >>> True
+
+        Use bulk update by passing a list of Records:
+
+        >>> devices = nb.dcim.devices.all()
+        >>> for d in devices:
+        >>>     d.name = d.name+'-test'
+        >>> nb.dcim.devices.update(devices)
+        >>> True
+        """
+        series = []
+        if not isinstance(objects, list):
+            raise ValueError('Objects passed must be list[dict|Record] '
+                             '- was ' + type(objects))
+        for o in objects:
+            if isinstance(o, Record):
+                data = o.updates()
+                if data:
+                    data['id'] = o.id
+                    series.append(data)
+            elif isinstance(o, dict):
+                if 'id' not in o:
+                    raise ValueError('id is missing from object: ' + str(o))
+                series.append(o)
+            else:
+                raise ValueError('Object passed must be dict|Record '
+                                 '- was ' + type(objects))
+        req = Request(
+            base=self.url,
+            token=self.token,
+            session_key=self.session_key,
+            http_session=self.api.http_session,
+        ).patch(series)
+
+        if isinstance(req, list):
+            return [self.return_obj(i, self.api, self) for i in req]
+        return self.return_obj(req, self.api, self)
+
     def choices(self):
         """ Returns all choices from the endpoint.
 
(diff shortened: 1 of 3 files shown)


Dimaqa:
Thats probably a bad way to implement bulk update and delete. The idea suggested by @zachmoody was to modify QuerySet class to get something like:
```
devices  = nb.dcim.devices.filter(status='active')
for device in devices:
    device.status = 'inventory'
devices.update()
devices.delete()
```
Or maybe even implement django usage:
```
devices  = nb.dcim.devices.filter(status='active')
devices.update(status='inventory')
devices.delete()
```
with 
```
for device in devices:
    device.status = 'inventory'
devices.save()
```

steffenschumacher:
> Thats probably a bad way to implement bulk update and delete.

uhm is it that you’d *also* like to be able to do this on querysets, or is there some specific bad thing in doing this from the endpoints?
I can think of usecases where I’d bulk update/delete records which are not in querysets, and how would I do that if not directly on the endpoints?

zachmoody:
> uhm is it that you’d _also_ like to be able to do this on querysets, or is there some specific bad thing in doing this from the endpoints? I can think of usecases where I’d bulk update/delete records which are not in querysets, and how would I do that if not directly on the endpoints? Eg. Deleting all interfaces of a virtual chassis, I think can be done in one step once you have all the interfaces, but they won’t be available in one recordset..

Hey, thanks for the PR. Apologies for not being able to comment sooner, but it just comes down to a UX choice. The idea behind implementing `RecordSet` was to allow for a little more fluid experience since you're able to lookup and update (or delete) in one call. E.g. `nb.dcim.devices.filter(tenant="blah").delete()` or `nb.dcim.devices.filter(tenant="blah").update(cf_some_field="blah")`.

> I can think of usecases where I’d bulk update/delete records which are not in querysets, and how would I do that if not directly on the endpoints?

Admittedly, I haven't been able to play around with the bulk action features in the new version, so I'm not up-to-speed on the details. The reason we haven't already implemented something like this is because you'd usually need an ID before being able to perform a bulk actions. Which means doing a lookup. Can bulk actions be done by something other than ID?

steffenschumacher:
> Hey, thanks for the PR. Apologies for not being able to comment sooner, but it just comes down to a UX choice. The idea behind implementing `RecordSet` was to allow for a little more fluid experience since you're able to lookup and update (or delete) in one call. E.g. `nb.dcim.devices.filter(tenant="blah").delete()` or `nb.dcim.devices.filter(tenant="blah").update(cf_some_field="blah")`.
Right, that's a valid point, so I'll add that option also - but I still think working directly on the endpoints can be warranted in certain cases - some which I mention in the docs:
* When you want to update/delete a subset of a `RecordSet`, because you need to filter on a something in `custom_fields` or on something else which the API doesn't (yet) support filtering on.
* When you want to consolidate/batch update/delete requests for a series of `RecordSet` - my own example is working with `VirtualChassis` which I use for stacked switches - when deleting or updating `Interface` for eg. 8 switches, then they are all fetched into separate `RecordSet` by filtering on the `device_id` of each of the virtual chassis members, however I can batch all of the update/delete operations of the combined lot of interfaces together into one API request, if using the endpoint. In fact I think it might make sense to allow the `endpoint.delete()` to accept a list of `RecordSet` objects for convenience?

> Admittedly, I haven't been able to play around with the bulk action features in the new version, so I'm not up-to-speed on the details. The reason we haven't already implemented something like this is because you'd usually need an ID before being able to perform a bulk actions. Which means doing a lookup. Can bulk actions be done by something other than ID?
Nope - you need the IDs, however that's not my point as described above.

I'll try to see if I can make time for adding the delete/update to `RecordSet` as discussed, but behind the scenes I think I'd just invoke them on the endpoint implementation I already did - I assume that would be ok?

zachmoody:
> But actually your example nb.dcim.devices.filter(tenant="blah").update(cf_some_field="blah") triggered me - how would I update custom fields?
Would I allow passing a new custom_fields dict? Would I check if the kwarg existed in custom_fields or both?
In the above you have used the cf_ prefix, but I guess I shouldn't put much into that or?
My approach would probably be to allow both (full dict and individual kwargs) - your thoughts?

haha, sorry, take that with a grain of salt. It was a totally contrived example, but yeah, would probably go with kwargs there and if someone wanted to pass a dict they can use `**` unpacking. `custom_fields` would still be a dict. Curious how the bulk API handles custom field updates? Like does passing an object there overwrite or merge?

Fair points on working with collections of RecordSets and subsets that can't be filtered from the NetBox API. I don't think this is a bad way of solving those use-cases, just one of a few you only get one shot at implementing :smile:. If we were to stay in keeping with the django ORM, it would be nice to deal with collections of RecordSets with unions (e.g. `combined = record_set_1 | record_set_2`). But applying that approach to the other issue is something I'm unsure I want to tackle anytime soon just due to the complexity in implementing `.filter()` on RecordSets.

PR #409, opened 2021-10-17
Fixes #316 bulk update/delete on both Endpoint and RecordSet

### LINKED ISSUE #316: Bulk delete and update support

steffenschumacher:
SoW: Add support for bulk updates and delete on Endpoint and RecordSet:
* Endpoint.update accepts one of:
  * list of dicts with id & updated attributes for low level access to the API
  * list of modified Records - unmodified Records are *not* updated and only modified attributes are sent
* Endpoint.delete accepts one of:
  * list of ids (int or numeric strings)
  * list of Records
  * RecordSet
* RecordSet.update accepts kwargs which are set onto its Records and then modifications are passed to its Endpoint
* RecordSet.delete does not accept arguments, and simply leverages the Endpoint.delete method
* Tests added:
  * Endpoint.delete: delete with {ids, Records, RecordSet}
  * Endpoint.update: update with {json, Records}. (list of Records & RecordSet is processed identically)
  * RecordSet.delete: delete (without args)
  * RecordSet.update: update with kwargs
* Added code is _blacked_ :)

### DIFF
--- a/pynetbox/core/endpoint.py
+++ b/pynetbox/core/endpoint.py
@@ -312,6 +312,129 @@ class Endpoint(object):
             return [self.return_obj(i, self.api, self) for i in req]
         return self.return_obj(req, self.api, self)
 
+    def update(self, objects):
+        r"""Bulk updates existing objects on an endpoint.
+
+        Allows for bulk updating of existing objects on an endpoint.
+        Objects is a list whic contain either json/dicts or Record
+        derived objects, which contain the updates to apply.
+        If json/dicts are used, then the id of the object *must* be
+        included
+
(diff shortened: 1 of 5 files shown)


zachmoody:
:bow: tyvm for the PR! Nice work!

Issue #412, opened 2021-11-08
Support manual pagination via offset


steffenschumacher:
TL;DR; to leverage pagination in the NB APIs, refactor code to also accept ```offset``` in the API vs handling offset internally as currently done.

Usecase: when building web apps on top of NB it becomes relevant to be able to control pagination in the requests to the APIs, to avoid redundant fetches. 

Proposal: Refactor code to accept ```offset``` in ```all()``` and ```filter()``` while still maintaining the existing behaviour, where if ```offset``` is not passed,```Request.get()``` will automatically fetch results using pagination using the current implementation.

Additionally: if I do a PR for this, could I then also include #386 in the fix?

Issue #386, opened 2021-06-24
Call filter() without filter raises an error.


BegBlev:
In my code, I call `filter()` method with dynamycally built filter `myfilter`.
`nb.dcim.device_types.filter(**myfilter)`

Sometimes it happens that `myfilter` is empty `{}` since I didn't specify any filters on my CLI arguments.
In that case, pynetbox raises an error.
`ValueError: filter must be passed kwargs. Perhaps use all() instead.`

This behaviour comes from code in core/endpoint.py
```
    def filter(self, *args, **kwargs):
    ...
        if not kwargs:
            raise ValueError("filter must be passed kwargs. Perhaps use all() instead.")
```

I could modify my code to test if `myfilter` is empty and then, call `all()` instead of `filter()` but this would be extra useless code.
In my opinion, calling filter with 'no filter' should not be an error this is just a generalized case. We could even imagine that `all()` method is only an 'alias' to `filter()` with no filter defined.
Actually, `all()` and `filter()` functions are very similar.
```
    def all(self, limit=0):
        ...
        req = Request(
            base="{}/".format(self.url),
            token=self.token,
            session_key=self.session_key,
            http_session=self.api.http_session,
            threading=self.api.threading,
            limit=limit,
        )
```
```
    def filter(self, *args, **kwargs):
        ...
        req = Request(
            filters=kwargs,
            base=self.url,
            token=self.token,
            session_key=self.session_key,
            http_session=self.api.http_session,
            threading=self.api.threading,
            limit=kwargs.get("limit", 0),
        )
```

The difference comes from the fact that `all()` method calls `Request` without setting `filters` argument.
In core/query.py we can see that the default value for `filters` parameter of `Request` constructor is `None`.
```
class Request(object):
    ...

    def __init__(
        self,
        base,
        http_session,
        filters=None,
        limit=None,
        key=None,
        token=None,
        private_key=None,
        session_key=None,
        threading=False,
    ):
```
This confirms that `filters` set to empty is only a special case of the generalized one.

Hence, I would suggest 2 modifications:
1. allow 'empty' filter for the `filter()` method
1. simplify `all()` method by just calling `filter()` with no filter parameter.

If those changes are approved, I can provide the pull request.

PR #413, opened 2021-11-08
Fixes manual pagination (#412) and argless filter (#386)

steffenschumacher:
SoW: Allow manual pagination and no-filter calls to ```Endpoint.filter()``` without breaking existing behaviour

- Endpoint.all() now also accepts ```offset``` which can be used with positive limit values - if not set, it is preexisting behaviour.
- Endpoint.filter() now also accepts ```offset``` which can be used with positive limit values - if not set, it is preexisting behaviour.
- Endpoint.filter() now accepts no filter args, which in practice means identical request as to Endpoint.all()
- Tests added:
  - Request.get: Test manual get (test_query.test_get_manual_pagination)
  - Endpoint.all: Test illegal use of offset with limit=0
  - Endpoint.list: Test illegal use of offset with limit=0
- Tests removed:
  - Endpoint.filter: test_filter_empty_kwargs
  - Endpoint.filter: test_filter_reserved_kwargs
- Added code is blacked :)
- All tests checks out

### DIFF
--- a/pynetbox/core/endpoint.py
+++ b/pynetbox/core/endpoint.py
@@ -16,7 +16,7 @@ limitations under the License.
 from pynetbox.core.query import Request, RequestError
 from pynetbox.core.response import Record, RecordSet
 
-RESERVED_KWARGS = ("offset",)
+RESERVED_KWARGS = ()
 
 
 class Endpoint(object):
@@ -71,13 +71,14 @@ class Endpoint(object):
             ret = Record
         return ret
 
-    def all(self, limit=0):
+    def all(self, limit=0, offset=None):
         """Queries the 'ListView' of a given endpoint.
 
         Returns all objects from an endpoint.
 
         :arg int,optional limit: Overrides the max page size on
             paginated returns.
+        :arg int,optional offset: Overrides the offset on paginated returns.
 
         :Returns: A :py:class:`.RecordSet` object.
 
@@ -92,6 +93,8 @@ class Endpoint(object):
         test1-leaf3
         >>>
         """
+        if limit == 0 and offset is not None:
+            raise ValueError("offset requires a positive limit value")
         req = Request(
             base="{}/".format(self.url),
             token=self.token,
@@ -99,6 +102,7 @@ class Endpoint(object):
             http_session=self.api.http_session,
             threading=self.api.threading,
             limit=limit,
+            offset=offset,
         )
 
         return RecordSet(self, req)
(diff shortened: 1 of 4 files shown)


pynetbox/core/endpoint.py
```diff
@@ -16,7 +16,7 @@
 from pynetbox.core.query import Request, RequestError
 from pynetbox.core.response import Record, RecordSet
 
-RESERVED_KWARGS = ("offset",)
+RESERVED_KWARGS = ()
```
steffenschumacher, line 19: Removed to enable manual pagination

pynetbox/core/endpoint.py
```diff
@@ -231,22 +236,24 @@ def filter(self, *args, **kwargs):
         if args:
             kwargs.update({"q": args[0]})
 
-        if not kwargs:
```
steffenschumacher, line 234: Remove to support #386

pynetbox/core/endpoint.py
```diff
@@ -231,22 +236,24 @@ def filter(self, *args, **kwargs):
...
+        if limit == 0 and offset is not None:
+            raise ValueError("offset requires a positive limit value")
         req = Request(
             filters=kwargs,
             base=self.url,
             token=self.token,
             session_key=self.session_key,
             http_session=self.api.http_session,
             threading=self.api.threading,
-            limit=kwargs.get("limit", 0),
```
steffenschumacher, line 249: I didn't check this, but if limit were in kwargs, wouldn't it also be passed as ```filters```? I think I fixed that with my approach - not sure it matters..

pynetbox/core/query.py
```diff
@@ -151,7 +152,7 @@ def __init__(
                 string.
         """
         self.base = self.normalize_url(base)
-        self.filters = filters
+        self.filters = filters or None
```
steffenschumacher, line 155: This is just to ensure we pass ```None``` if filters is an empty dict (```{}```)

tests/unit/test_endpoint.py
```diff
@@ -22,21 +22,21 @@ def test_filter(self):
             test = test_obj.filter(test="test")
             self.assertEqual(len(test), 2)
 
-    def test_filter_empty_kwargs(self):
```
steffenschumacher, line 25: Neither this nor test_filter_reserved_kwargs makes sense with this PR, so I've lazily converted them to another type of test..

aleyval:
Hi ! Thanks for the fix, I definitely need this : when will it be merged ? ++

steffenschumacher:
> Hi ! Thanks for the fix, I definitely need this : when will it be merged ? ++

Dunno - possibly they are waiting for more PRs😕

zachmoody (approved):

PR #419, opened 2021-12-05
Use `display` field in `Record.__str__()` to support future models

### LINKED ISSUE #417: The `ipaddresses` of a Service return IP elements with an empty representation
```python
services = list(nb.ipam.services.filter(tag="node-exporter")
print(svc.ipaddresses)
# []
# prints an empty list
```

The list itself is not empty, but the `Record` object has an empty representation (which is very confusing and I was troubleshooting a completely bogus problem).

As of right now, I am using last version of pynetbox (6.2.0) and last version of NetBox (3.0.11). I noticed this "blank representation" bug with NetBox v3.0.2 and just upgraded, and I keep seeing this same behavior.

markkuleinio:
This PR supersedes #388 for clarity, and should fix #417, example:

```
>>> ss = list(netbox.ipam.services.all())
>>> ss[0].ipaddresses
[1.1.1.1/32]
>>>
```

See also: https://github.com/netbox-community/netbox/pull/5992 (= NetBox 2.11.0 added the `display` fields)

### DIFF
--- a/pynetbox/core/response.py
+++ b/pynetbox/core/response.py
@@ -313,7 +313,12 @@ class Record(object):
         return dict(self)[k]
 
     def __str__(self):
-        return getattr(self, "name", None) or getattr(self, "label", None) or ""
+        return (
+            getattr(self, "name", None)
+            or getattr(self, "label", None)
+            or getattr(self, "display", None)
+            or ""
+        )
 
     def __repr__(self):
         return str(self)
--- a/tests/fixtures/users/unknown_model.json
+++ b/tests/fixtures/users/unknown_model.json
@@ -1,0 +1,4 @@
+{
+    "id": 1,
+    "display": "Unknown object"
+}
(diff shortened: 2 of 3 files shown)


markkuleinio:
In addition to this PR, do you @zachmoody see us adding new custom models in `pynetbox` as they appear in NetBox? I mean, I could add a PR adding models like `extras.Tags` and so on, if desired. But if we are not going to add new models frequently, I'm not sure about the benefit of adding them at all (unless there is a need like with the model fields).

zachmoody (approved):

zachmoody:
> In addition to this PR, do you @zachmoody see us adding new custom models in `pynetbox` as they appear in NetBox? I mean, I could add a PR adding models like `extras.Tags` and so on, if desired. But if we are not going to add new models frequently, I'm not sure about the benefit of adding them at all (unless there is a need like with the model fields).

Models were always intended to just handle outliers. Kind of a judgement call on when to move that behavior into the parent class. I'd say if more than a few models behave similarly we should think about it. This totally make sense though if `display` is going to be the new human-friendly field going forward.

PR #423, opened 2021-12-11
Fixes #422: Adds support for HTTP 409 when allocation is unsuccessful

### LINKED ISSUE #422: NetBox 3.1.1+ returns HTTP 409 instead of 204 when available-ips/prefixes request cannot be fulfilled
In NetBox https://github.com/netbox-community/netbox/commit/7c14c0812b92507082116b760211555f1fc75a76 the unsuccessful IP or prefix allocation requests with `available-ips` or `available-prefixes` API child endpoints will return HTTP status code 409 instead of 204. This will require change in `pynetbox` to raise `AllocationError` correctly.

markkuleinio:
HTTP status code 204 (No Content) for a POST request previously raised `AllocationError`, in practice in situations where `/prefixes/xx/available-ips/` or `/prefixes/xx/available-prefixes/` was attempted in NetBox to allocate new IP(s) or prefix(es) but there was no space for the allocation.

As mentioned in #422 NetBox 3.1.1 will return 409 (Conflict) instead in that situation, so this PR adds support for that: both 204 and 409 for a POST request will raise `AllocationError` in `pynetbox`.

AFAIK NetBox will not return 409 for a POST request in any other situations.

`tests.util.Response` class was also changed to make it possible for `AllocationError` to parse required fields from the patched response.

Fixes #422

### DIFF
--- a/pynetbox/core/query.py
+++ b/pynetbox/core/query.py
@@ -268,7 +268,7 @@ class Request(object):
             url_override or self.url, headers=headers, params=params, json=data
         )
 
-        if req.status_code == 204 and verb == "post":
+        if req.status_code in [204, 409] and verb == "post":
             raise AllocationError(req)
         if verb == "delete":
             if req.ok:
--- a/pynetbox/core/query.py
+++ b/pynetbox/core/query.py
@@ -73,7 +73,8 @@ class AllocationError(Exception):
     """Allocation Exception
 
     Used with available-ips/available-prefixes when there is no
-    room for allocation and NetBox returns 204 No Content.
+    room for allocation and NetBox returns 204 No Content (before
+    NetBox 3.1.1) or 409 Conflict (since NetBox 3.1.1+).
     """
 
     def __init__(self, message):
--- a/tests/unit/test_query.py
+++ b/tests/unit/test_query.py
@@ -2,14 +2,12 @@ import unittest
 
 import six
 
-import pynetbox
-from pynetbox.core.query import AllocationError, Request
-from ..util import Response
+from pynetbox.core.query import Request
 
 if six.PY3:
-    from unittest.mock import patch, Mock, call
+    from unittest.mock import Mock, call
 else:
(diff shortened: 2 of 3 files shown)


pynetbox/core/query.py
```diff
@@ -268,7 +269,7 @@ def _make_call(self, verb="get", url_override=None, add_params=None, data=None):
             url_override or self.url, headers=headers, params=params, json=data
         )
 
-        if req.status_code == 204 and verb == "post":
+        if req.status_code in [204, 409] and verb == "post":
```
zachmoody, line 272: This'll work, but I think I want to refactor this a bit at some point to not handle errors here.

tests/unit/test_query.py
```diff
@@ -85,3 +87,29 @@ def test_get_manual_pagination(self):
...
+        "requests.sessions.Session.get",
+        return_value=Response(fixture="ipam/prefix.json"),
+    )
+    @patch(
+        "requests.sessions.Session.post", return_value=Response(status_code=409),
+    )
+    def test_409_case(self, _, __):
+        nb = pynetbox.api("http://localhost:8000/")
+        with self.assertRaises(AllocationError):
+            _ = nb.ipam.prefixes.get(1).available_ips.create()
```
zachmoody, line 115: Trying to get away from the tests that rely on fixtures. Given I want to refactor `_make_call`, I can probably manage to test this in the new setup.

markkuleinio:
@zachmoody have you considered merging this in and doing the possible code reorgs later separately? I mean, this *is* a major blocker for going to the latest NetBox 3.1.1. Deploying my own fork to prod CI/CD next week is not tempting...

zachmoody:
Yeah, totally, that's mostly where I was coming from. Just thought it'd be easier to drop the tests now instead of remembering to later. I'll merge tomorrow, if you have a chance to push a commit removing them before then, great. Nbd if not.

markkuleinio:
👍

PR #426, opened 2021-12-23
Do not skip `plugins` in endpoints URL

### LINKED ISSUE #425: Endpoint.url and Record.url doesn’t match
Hi.

I’m trying to script something around the `netbox_dns` plugin API and I’m unable to query zone records (at `/api/plugins/netbox_dns/{id}/record`).

I’ve tried using `DetailEndpoint`, but the class is getting its base URL from `parent_obj.endpoint.url` which has the wrong value:

```
>>> z = nb.plugins.netbox_dns.zones.get(1)
>>> z.endpoint.url
'https://netbox.br0.fr/api/netbox-dns/zones'
```

`Record.url` seems correct tho:
```
>>> z.url
'https://netbox.br0.fr/api/plugins/netbox-dns/zones/1/'
```

## Details

* netbox: 3.1.1
* pynetbox: 6.4.1
* plugins: only [netbox-dns](https://github.com/auroraresearchlab/netbox-dns) (0.7.0)

johanfleury:
Fix an issue introduced by #324.

I’m not sure whether or not this re-introduce the issue reported in #302 but I don’t know how to reproduce it, so please be careful before merging this.

---

Fixes #425

### DIFF
--- a/pynetbox/core/response.py
+++ b/pynetbox/core/response.py
@@ -399,8 +399,8 @@ class Record(object):
             url_path = url_path[len(extra_path) :]
         split_url_path = url_path.split("/")
         if split_url_path[2] == "plugins":
-            # Skip plugins from the path
-            app, name = split_url_path[3:5]
+            app = "plugins/{}".format(split_url_path[3])
+            name = split_url_path[4]
         else:
             app, name = split_url_path[2:4]
         return getattr(pynetbox.core.app.App(self.api, app), name)


johanfleury:
I tried to run pytest locally but had an issue with fixtures, the tests introduced by #324 in `test_response.py` are passing tho.

jcollie:
Adding this patch to my local copy of pynetbox lets API calls that delete or update plugin objects work.

PR #430, opened 2021-12-28
Allow use of plugin-secrets-store with Netbox 3.x

uedvt359:
currently, `nb.secrets` tries to access Netbox 2.x's built-in secrets store. In version 3.x, a plugin must be used instead; this patch
redirects `nb.secrets` to the plugin provided to `pynetbox.api(secrets_provider=...)`, so scripts using pynetbox can easily
be written to support both Netbox 2.x and 3.x.
All that is needed is to give the desired secret-store-plugin to pynetbox.api(), e.g.
`nb = pynetbox.api(..., secrets_provider="netbox_secretstore")`
Later, the user can use `nb.secrets.secrets.get(...)` in exactly the same way on either netbox version.

This required the following secondary change:

because pynetbox translates underscores to dashes when accessing attributes, it is not possible to use plugins that have actual
underscores in their URLs. For example, `nb.plugins.plugin_with_underscores` actually queries `plugin-with-underscores` (note the dashes) instead.
This adds a way to access those URLs, by using the following syntax: `nb.plugins['plugin_with_underscores']`

Here's an example how that would look:

```py
import pynetbox

nb = pynetbox.api(
    "https://netbox.example",
    token="...",
    private_key_file="private.key",
    secrets_provider="netbox_secretstore"
)
# automatically detect the correct secretstore based on the secrets_provider:
secret = nb.secrets.secrets.get(device="device01.example", name="username")
print("the username is", secret)

# explicitly accessing the secretstore plugin:
# note that this still needs secrets_provider set, otherwise fetching the session key will fail
nb.plugins['netbox_secretstore'].secrets.get(device="device01.example", name="username")
print("the username is", secret)
# (you wouldn't use this for secrets, but other plugins might profit from this syntax)
```

### DIFF
--- a/pynetbox/core/api.py
+++ b/pynetbox/core/api.py
@@ -32,7 +32,7 @@ class Api(object):
         * dcim
         * ipam
         * circuits
-        * secrets (on NetBox 2.11 and older)
+        * secrets
         * tenancy
         * extras
         * virtualization
@@ -53,8 +53,9 @@ class Api(object):
         wish to connect to.
     :param str token: Your NetBox token.
     :param str,optional private_key_file: The path to your private
-        key file. (Usable only on NetBox 2.11 and older)
-    :param str,optional private_key: Your private key. (Usable only on NetBox 2.11 and older)
+        key file.
+    :param str,optional private_key: Your private key.
+    :param str,optional secrets_provider: Plugin to use instead of built-in secrets.
     :param bool,optional threading: Set to True to use threading in ``.all()``
         and ``.filter()`` requests.
     :raises ValueError: If *private_key* and *private_key_file* are both
@@ -78,6 +79,7 @@ class Api(object):
         private_key=None,
         private_key_file=None,
         threading=False,
+        secrets_provider=None,
     ):
         if private_key and private_key_file:
             raise ValueError(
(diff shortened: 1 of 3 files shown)


uedvt359:
rebased on top of master.

arthanson:
I'd like to keep plugins out of the code pynetbox - it opens up a can of worms as to which plugins we support.  In the future could potentially look at an extension mechanism of some kind.  Closing now for cleanup.

marinus81:
@arthanson - I understand the issue you're trying to avoid. Unfortunately since core functionality was "extracted" into an (optional) plugin this is kind of hanging mid-air. Can you think of any (future) workaround for applications relying on this functionality of pynetbox after upgrade to 3.x+?

uedvt359:
Well, this is really unfortunate. I understand your concern about which plugins to support, but this shouldn't be the can of worms you think it is.

This patch hooks into a lot of pre-existing code, which other pluigins don't. And it does so in a very generic way. First, it doesn't hard-code any specific plugin implementation for the secretstore functionality, so we are not really playing favourites here. Secondly, commit ec5cbd0bc7b7ac2bb590062760d90290909ee0f7 adds a generic way to access plugin URLs (with underscores), so there should be no further support requests for normal plugins.

In other words, the secret store plugin is the only special one, since it replaces core functionality, while other plugins don't.

I really hope you can re-consider this decision.

Issue #392, opened 2021-07-02
Support get()/all()/filter() to return model class instead of Record class


lordofire:
When we do certain resource type get()/all()/filter() operations, the returned objects have class of Record. However, under the situation that each of the above methods are read under certain endpoint, it looks pretty straightforward to return the class defined in the model instead of base class Record. Also, I saw there is a way to pass in `model` [here](https://github.com/netbox-community/pynetbox/blob/master/pynetbox/core/endpoint.py#L55), but seems not used.

steffenschumacher:
I guess this would need someone to create all the models used in NB, and keep these up to date all the time?
Also you wouldn't have the custom fields, since these would all be depending on each install - I think it might be better to somehow read the openapi spec + somehow get custom fields and have a script compile/generate the model classes which could be loaded and used if the name matches - but this is a huge effort I think?

zachmoody:
Yeah, the approach I went with for pynetbox in the beginning was to have `Record` cover as many use-cases as was reasonable without adding a ton of conditionals for outliers that were a little different from the majority. Those exceptions were addressed with models. I wanted to limit the number we had to write. Using the openapi spec's been brought up before, but there's been a few times it hasn't accurately represented the API. That was a long time ago though, I don't know if it's any better.

johanfleury:
@zachmoody I was looking at a way to use external models and I think I came up with something that could fix this issue, but it uses a “global variable”, which you might not like:

```diff
diff --git a/pynetbox/__init__.py b/pynetbox/__init__.py
index 0e1f510..abbbf04 100644
--- a/pynetbox/__init__.py
+++ b/pynetbox/__init__.py
@@ -2,6 +2,7 @@ from pkg_resources import get_distribution, DistributionNotFound

 from pynetbox.core.query import RequestError, AllocationError, ContentError
 from pynetbox.core.api import Api as api
+from pynetbox.core.app import register_model

 try:
     __version__ = get_distribution(__name__).version
diff --git a/pynetbox/core/app.py b/pynetbox/core/app.py
index 797eb17..9991c06 100644
--- a/pynetbox/core/app.py
+++ b/pynetbox/core/app.py
@@ -25,6 +25,16 @@ from pynetbox.models import (
     wireless,
 )

+MODELS = {
+    "dcim": dcim,
+    "ipam": ipam,
+    "circuits": circuits,
+    "virtualization": virtualization,
+    "extras": extras,
+    "users": users,
+    "wireless": wireless,
[...]

PR #433, opened 2022-01-03
Allow registration of external models

johanfleury:
Implements #392.

### DIFF
--- a/docs/advanced.rst
+++ b/docs/advanced.rst
@@ -1,3 +1,7 @@
+==============
+Advanced usage
+==============
+
 Custom Sessions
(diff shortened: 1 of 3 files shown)


docs/advanced.rst
```diff
@@ -1,3 +1,7 @@
+==============
```
johanfleury, line 1: I have a PR underway to rework the doc, this file is currently not included in any toctree and sphinx complains with:
```
WARNING: document isn't included in any toctree
```

abhi1693:
This is not needed since the introduction of `Plugins` app.

PR #439, opened 2022-01-22
Fix `VirtualChassis.__str__()`, make `VirtualChassis.master` a `Devices` object

### LINKED ISSUE #328: AttributeError: NoneType of VirtualChassis str method
Hi,

VirtualChassis can be without master, and [this](https://github.com/digitalocean/pynetbox/blob/v5.1.0/pynetbox/models/dcim.py#L95) method raised with AttributeError. Maybe just the attribute __self.name__?
```python
class VirtualChassis(Record):
    def __str__(self):
        return self.master.display_name
```

Traceback:
```
  File "test.py", line 456, in discovery_hosts
    print(virtual_chassis)
  File "./lib/python2.7/site-packages/pynetbox/core/response.py", line 217, in __repr__
    return str(self)
  File "./lib/python2.7/site-packages/pynetbox/models/dcim.py", line 95, in __str__
    return self.master.display_name
AttributeError: 'NoneType' object has no attribute 'display_name'
```

Used pynetbox v5.1.0 with netbox v2.10.4

### LINKED ISSUE #437: VirtualChassis DCIM use of display_name
When trying to do `str(foo)` where foo represents a Virtual Chassis, an exception is thrown:

```
  File "script.py", line X, in <module>
    if str(foo):
  File "/usr/local/lib/python3.6/site-packages/pynetbox/models/dcim.py", line 156, in __str__
    return self.master.display_name
  File "/usr/local/lib/python3.6/site-packages/pynetbox/core/response.py", line 298, in __getattr__
    raise AttributeError('object has no attribute "{}"'.format(k))
AttributeError: object has no attribute "display_name
```

I believe `display_name` has been deprecated, and so this should instead be either `.name` or `.display`.

markkuleinio:
Fixes #328 (VirtualChassis without master)
Fixes #437 (NetBox 2.11+ does not have `display_name` field anymore)

I don't see a reason why a virtual chassis string representation should consist of the master device name when the virtual chassis has its own name, thus the default behaviour of `Record.__str__()` is enough. **Note: This is a change of behaviour in `pynetbox`.** If that is not desired, then we can check for the master device existence and return the VC name only if master is not found.

I see `VirtualChassis.name` was added and master device was made optional in NetBox 2.9: https://netbox.readthedocs.io/en/stable/release-notes/version-2.9/#named-virtual-chassis-2018

Additionally the `master` attribute is made a `Devices` object.

### DIFF
--- a/pynetbox/models/dcim.py
+++ b/pynetbox/models/dcim.py
@@ -152,8 +152,7 @@ class RackReservations(Record):
 
 
 class VirtualChassis(Record):
-    def __str__(self):
-        return self.master.display_name
+    master = Devices
 
 
 class RUs(Record):


markkuleinio:
I see there are `tests/fixtures/dcim/virtual_chassis_*.json` but I didn't find where they are used. Also, the tests pass even though those fixture data don't have `name`/`label`/`display` attribute (that `Record.__str__()` use). But I think @zachmoody said earlier something about changing the whole test logic, so let me know if some tests are still desired in the old style.

zachmoody:
> I see there are `tests/fixtures/dcim/virtual_chassis_*.json` but I didn't find where they are used. Also, the tests pass even though those fixture data don't have `name`/`label`/`display` attribute (that `Record.__str__()` use). But I think @zachmoody said earlier something about changing the whole test logic, so let me know if some tests are still desired in the old style.

:+1:  part of the reason I'd like to get the old fixture based tests moved over to the integration testing.

Issue #434, opened 2022-01-04
/path not /trace on front and rear ports


marcelb98:
On front and rear ports the API provides `/path` and not `/trace` which is currently usable by the `.trace()` function on TracableRecords (which front and rear ports seem to be at the moment).

Calling the `.trace()` function on rear ports actually leads to an AttributeError (as there is no `/trace` provided by the Netbox API).

 Maybe this changed from some older versions of Netbox…

markkuleinio:
NetBox 2.10 release notes (https://netbox.readthedocs.io/en/stable/release-notes/version-2.10/) say:

> dcim.FrontPort:
>   Replaced the /trace/ endpoint with /paths/, which returns a list of cable paths

> dcim.RearPort:
>    Replaced the /trace/ endpoint with /paths/, which returns a list of cable paths

It also says:

> circuits.CircuitTermination:
>    Added the /trace/ endpoint

> dcim.PowerFeed:
>    Added the /trace/ endpoint

In `pynetbox`, these all models inherit from `TraceableRecord` (that defines the `trace()` method):

- Interfaces
- PowerOutlets
- PowerPorts
- ConsolePorts
- ConsoleServerPorts
- FrontPorts
- RearPorts

Maybe just copying `TraceableRecord.trace()` to `TraceableRecord.paths()` (with `/paths` request path in the code instead of `/trace`) (in `pynetbox/models/dcim.py`) could do? @marcelb98 are you able to test it with your data?

marcelb98:
Just copying the `trace()` to `paths()` and changing the requested URL doesn't work, as the `/paths` endpoint returns a different datastructure. While the `/trace` for interfaces returns a list of `(Endpoint, Cable, Endpoint)` tuples, the `/paths` returns a list of related paths, each with an origin and destination Endpoint and the path from origin to endpoint as a list of Interfaces, Front/RearPorts and Cables.

I've implemented a `paths()` function in PR #440.

PR #440, opened 2022-01-23
add cable trace (pathing) for front-ports and rear-ports

marcelb98:
The Netbox API has no `/trace` endpoint for front-ports and rear-ports, but a `/paths` endpoint (see issue #434).
This PR implements the `/paths` endpoint for pynetbox.

### DIFF
--- a/pynetbox/models/dcim.py
+++ b/pynetbox/models/dcim.py
@@ -64,6 +64,55 @@ class TraceableRecord(Record):
         return ret
 
 
+class PathableRecord(Record):
+    def paths(self):
+        req = Request(
+            key=str(self.id) + "/paths",
+            base=self.endpoint.url,
+            token=self.api.token,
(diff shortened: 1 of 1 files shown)


srfwx:
Hi,
I've tested your code. It doesn't work for me when tracing an open path.
I think a condition is missing in case there is no origin/destination port.

` origin_endpoint = "/".join(
                    urlsplit(origin["url"])
                        .path[len(urlsplit(self.api.base_url).path):]
                        .split("/")[1:3]
                ) if origin else None
 `
 `
             destination_endpoint = "/".join(
                urlsplit(destination["url"])
                    .path[len(urlsplit(self.api.base_url).path):]
                    .split("/")[1:3]
            ) if destination else None
            `

arthanson:
@marcelb98 can you please provide a test case for this?

arthanson:
Closing as no response.

PR #441, opened 2022-01-30
Upgrade Black to the first non-beta release

markkuleinio:
Since Black has been selected to be the code formatter for `pynetbox`, let's upgrade it to the first non-beta release.

`requirements-dev.txt` has been set to `black~=22.0` as per the [Black Stability Policy](https://black.readthedocs.io/en/stable/the_black_code_style/index.html#stability-policy).

PR #444, opened 2022-02-05
ci: update the versions of netbox in the test suite

### LINKED ISSUE #438: Adding recent NetBox versions in integration tests
The integration tests implemented last year in `pynetbox` only test against NetBox 2.7, 2.8, 2.9 and 2.10. Currently the three latest NetBox releases are 2.11, 3.0 and 3.1.

Would @raddessi or someone else be willing to update the integration tests for recent NetBox versions? Or was there some intention to rewrite the tests somehow else?

Btw I had a quick Twitter poll in December about the NetBox versions that people mostly use 😊

image

raddessi:
* Bump the versions of netbox we test against to the latest 3 major/minor netbox releases: ~2.10~ 2.11, 3.0, and 3.1.
* Added the netbox version in the test suite to the github matrix strategy to speed up the test suite.

Closes #438

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -4,7 +4,7 @@ import pytest
 from packaging import version
 
 
-DEFAULT_NETBOX_VERSIONS = "2.10, 3.0, 3.1"
+DEFAULT_NETBOX_VERSIONS = "2.11, 3.0, 3.1"
 
 
 def pytest_addoption(parser):
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -4,7 +4,7 @@ import pytest
 from packaging import version
 
 
-DEFAULT_NETBOX_VERSIONS = "2.7, 2.8, 2.9, 2.10"
+DEFAULT_NETBOX_VERSIONS = "2.10, 3.0, 3.1"
 
 
 def pytest_addoption(parser):
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -23,7 +23,7 @@ def get_netbox_docker_version_tag(netbox_version):
 
     Returns:
         str: The release tag for the netbox-docker repo that should be able to build
-            the rgequested version of netbox.
+            the requested version of netbox.
 
     """
     major, minor = netbox_version.major, netbox_version.minor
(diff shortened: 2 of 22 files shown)


raddessi:
Ah I see the black update conflicts with https://github.com/netbox-community/pynetbox/pull/441, thank you for that mark, I'll revert that part in this one.

tests/conftest.py
```diff
@@ -4,7 +4,7 @@
 from packaging import version
 
 
-DEFAULT_NETBOX_VERSIONS = "2.7, 2.8, 2.9, 2.10"
+DEFAULT_NETBOX_VERSIONS = "2.10, 3.0, 3.1"
```
raddessi, line 7: Are these the only versions we want to test against? We could leave more in here and with the github matrix strategy it wouldn't really slow down the test suite runs on github, only the local runs if someone was to run the whole suite.

raddessi (commented):
Only one question

tests/integration/conftest.py
```diff
@@ -251,6 +252,10 @@ def docker_compose_file(pytestconfig, netbox_docker_repo_dirpaths):
                         "netboxcommunity/netbox:v%s" % netbox_version
                     )
 
+                if service_name == "netbox":
+                    # ensure the netbox container listens on a random port
+                    new_services[new_service_name]["ports"] = ["8080"]
+
```
zachmoody, line 258: :bow: this was foiling my attempts to get this going a few weeks ago :laughing:

tests/conftest.py
zachmoody: What happened to 2.11? But otherwise, I'm cool with sticking to latest three major releases. Probably don't want to be unnecessarily rough on free resources. :)

zachmoody (changes requested):
Nice, thanks for updating these. With #441's merging, would you mind re-running black too?

tests/integration/conftest.py
raddessi: Yeah, they apparently removed all default `port` listener statements from the default config so that people could have more control. Makes sense but I wish the project was a little more backwards compatible :)

raddessi:
> would you mind re-running black too?

Sweet! I'd love to :)

> What happened to 2.11?

Ha, I forgot 2.11 was a thing. Sorry about that, good catch. Will fix AM.

raddessi:
Looks good to me now.

PR #445, opened 2022-02-07
Use devices as variable for example calls to nb.dcim.devices

berendt:
When calling nb.dcim.devices in the examples, devices
should be used as a variable.

Signed-off-by: Christian Berendt <berendt@osism.tech>

### DIFF
--- a/pynetbox/core/endpoint.py
+++ b/pynetbox/core/endpoint.py
@@ -225,7 +225,7 @@ class Endpoint(object):
         Passing a list as a named argument adds multiple filters of the
         same value.
 
-        >>> device = nb.dcim.devices.filter(role=['leaf-switch', 'spine-switch'])
+        >>> devices = nb.dcim.devices.filter(role=['leaf-switch', 'spine-switch'])
         >>> for device in devices:
         ...     print(device.name)
         ...


pynetbox/models/dcim.py
```diff
@@ -106,8 +106,8 @@ def napalm(self):
 
         :Examples:
 
-        >>> device = nb.ipam.devices.get(123)
-        >>> device.napalm.list(method='get_facts')
+        >>> devices = nb.ipam.devices.get(123)
```
markkuleinio, line 109: `Endpoint.get()` returns one object only

pynetbox/core/endpoint.py
```diff
@@ -225,7 +225,7 @@ def filter(self, *args, **kwargs):
         Passing a list as a named argument adds multiple filters of the
         same value.
 
-        >>> device = nb.dcim.devices.filter(role=['leaf-switch', 'spine-switch'])
+        >>> devices = nb.dcim.devices.filter(role=['leaf-switch', 'spine-switch'])
```
markkuleinio, line 228: 👍

markkuleinio (approved):

markkuleinio (approved):

PR #446, opened 2022-02-07
ci: test against the currently supported python versions

raddessi:
While working on https://github.com/netbox-community/pynetbox/pull/444 I noticed that python 3.6 was still being tested here while official support of that python version [was dropped in netbox v3.0.0](https://netbox.readthedocs.io/en/stable/#supported-python-versions). 

Once https://github.com/netbox-community/pynetbox/pull/444 is merged I will rebase this one.

### DIFF
--- a/.github/workflows/py3.yml
+++ b/.github/workflows/py3.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python: [3.6, 3.8]
+        python: [3.7, 3.8, 3.9]
 
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/py3.yml
+++ b/.github/workflows/py3.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python: [3.7, 3.10]
+        python: ["3.7", "3.10"]
         netbox: ["2.11", "3.0", "3.1"]
 
     steps:


markkuleinio:
I'm sure there are lots of opinions on this. My take:
- If you are maintaining such old systems that only run EoL Python (like 3.6), you most probably won't be upgrading your packages just like that. At least you shouldn't do that! Thus, no harm done. Dropping testing support for 3.6 should be mentioned in the release notes though.
- The newest Python release (currently 3.10) should always be included.
- I don't know if we should test all versions all the time though, to save some of the free resources. Maybe just 3.7 and 3.10 (= the oldest and newest Python). And then shift the versions forward with Python lifecycle.

Btw https://package.wiki/pynetbox shows interesting data (a few months old but anyway) about how other packages pin the dependencies.

raddessi:
> If you are maintaining such old systems that only run EoL Python (like 3.6), you most probably won't be upgrading your packages just like that. At least you shouldn't do that! Thus, no harm done. Dropping testing support for 3.6 should be mentioned in the release notes though.

I agree.

> The newest Python release (currently 3.10) should always be included.

This is also a great point.

> I don't know if we should test all versions all the time though, to save some of the free resources. Maybe just 3.7 and 3.10 (= the oldest and newest Python). And then shift the versions forward with Python lifecycle.

As an alternative, if anyone is interested, the current `py3` workflow could be split up so that PRs only test with a subset of versions (oldest/newest like you said?) but then any merge to the `master` branch would test against all versions.

I'm personally not too worried about using CI time, I think it ends up better for end users if we can confirm compatibility across all versions. Of course there is always a trade off and you don't want to waste cycles if you are testing something that would not provide any useful information but this test suite is currently very fast (in my opinion). I'm just not sure how actually useful it would be to test middle versions, I'm not sure if they would catch anything the first and last would not.

PR #459, opened 2022-04-08
Fix pickling exception with custom errors

kingtong:
**Description**
I ran into an issue when using pynetbox with the `multiprocessing` module. Using a `Pool`, data is serialized between processes using `pickle`.

In my case, a 502 is returned from time to time by our instance of Netbox, resulting in an exception in pynetbox when building a `RequestError` in one of the processes of the `Pool`. An exception is raised during the unpickle (details below).

**To reproduce**
```python
import pickle

from requests import PreparedRequest, Response
from pynetbox import RequestError

if __name__ == '__main__':
    req = PreparedRequest()
    req.body = 'Bad Gateway'

    resp = Response()
    resp.status_code = 502
    resp.reason = 'Bad Gateway'
    resp.request = req

    pickle.loads(pickle.dumps(RequestError(resp)))
```

**Expected behaviour**
No error.

**Actual behaviour**
```
  File "/Users/eq01458/Development/tmp/python/snippet.py", line 16, in <module>
    pickle.loads(pickle.dumps(RequestError(resp)))
  File "/Users/eq01458/.pyenv/versions/3.10/lib/python3.10/site-packages/pynetbox/core/query.py", line 49, in __init__
    if req.status_code == 404:
AttributeError: 'str' object has no attribute 'status_code'
```

**Fix**
The fix proposed in this PR is:
* Pass along the same request/response object to the base exception constructor
* Implement the `__str__` to get the same behaviour as using the message argument for the base exception

### DIFF
--- a/pynetbox/core/query.py
+++ b/pynetbox/core/query.py
@@ -102,7 +102,9 @@ class ContentError(Exception):
         self.req = req
         self.request_body = req.request.body
         self.base = req.url
-        self.error = "The server returned invalid (non-json) data. Maybe not a NetBox server?"
+        self.error = (
+            "The server returned invalid (non-json) data. Maybe not a NetBox server?"
+        )
 
     def __str__(self):
         return self.error
--- a/pynetbox/core/query.py
+++ b/pynetbox/core/query.py
@@ -43,18 +43,16 @@ class RequestError(Exception):
 
     """
 
-    def __init__(self, message):
-        req = message
-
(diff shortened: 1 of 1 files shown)


zachmoody:
Makes sense, thanks for the PR! Happy to merge if you don't mind making the linter happy first. Thanks!

kingtong:
Thanks for the feedback. I've pushed the fix for the linter.

PR #460, opened 2022-04-12
Add retry on transient errors logic

kingtong:
From time to time I receive some transient errors from our Netbox instance and / or the proxy that sits between.

This PR adds the retry mechanism for transient errors.

### DIFF
--- a/pynetbox/core/api.py
+++ b/pynetbox/core/api.py
@@ -17,6 +17,7 @@ import sys
 
 import requests
 
+from pynetbox.core.query import RETRYABLE_TRANSIENT_MAX_RETRIES
 from pynetbox.core.query import Request
 from pynetbox.core.app import App, PluginsApp
 from pynetbox.core.response import Record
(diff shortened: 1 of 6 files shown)


zachmoody:
:wave:  The [docs](https://pynetbox.readthedocs.io/en/latest/advanced.html#timeouts) already have instructions on timeout handling.

zachmoody:
:wave:  I think you should be able to use the [custom sessions pattern](https://pynetbox.readthedocs.io/en/latest/advanced.html#custom-sessions) to handle retries as well.

markkuleinio:
Example:

```
import pynetbox
import requests
from requests.adapters import HTTPAdapter
from urllib3 import Retry

nb = pynetbox.api(YOUR_NETBOX_URL, YOUR_NETBOX_TOKEN)
retry = Retry(total=5, backoff_factor=1, status_forcelist=[429, 500, 502, 503, 504], allowed_methods=None)
nb.http_session = requests.Session()
nb.http_session.mount("http://", HTTPAdapter(max_retries=retry)) # change to https if used
```

Test with stopping the netbox service on the server so that nginx/apache will return "502 Bad Gateway" errors. See that a simple `nb.version` call will return at 30 seconds with `requests.exceptions.RetryError: HTTPConnectionPool(host='netbox-test.lein.io', port=80): Max retries exceeded with url: /api/ (Caused by ResponseError('too many 502 error responses'))`.

With `total=5` and `backoff_factor=1` the retries will happen at 0, 2, 4, 8 and 16 seconds (total = 30).

Docs: https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html#module-urllib3.util.retry

kingtong:
Thanks a lot for the feedback.
Indeed, relying on a custom session with the `Retry` will meet my need.
Closing the PR.

PR #461, opened 2022-04-18
CI: Add integration testing for NetBox 3.2

markkuleinio:
At the same time I dropped testing for 3.0, I'm thinking along the lines that testing for 3.1 should represent 3.0 well enough.

### DIFF
--- a/.github/workflows/py3.yml
+++ b/.github/workflows/py3.yml
@@ -13,7 +13,7 @@ jobs:
     strategy:
       matrix:
         python: ["3.7", "3.10"]
-        netbox: ["2.11", "3.0", "3.1"]
+        netbox: ["2.11", "3.1", "3.2"]
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


markkuleinio:
... except that I forgot that NetBox 3.2 will not run with Python 3.7, so some other refactoring in the GitHub workflow will be needed as well.

All checks were successful though, so I don't know what to make of it 😃 I guess the container packaging takes over the situation. So, **for what** is the Python version actually used in the GitHub workflow...?

raddessi:
Hmm that is a good question.. I thought I had it passing the python version down to the container but I may have skipped that.. how would you guys like this to work?

raddessi:
Yeah you are correct, the netbox-docker container has the python version already set.. I think the different python versions here are valid only for the unit tests. Maybe we should split up the unit and integration tests in to two separate runs, with the units still getting the python version matrix and the containerized tests not?

markkuleinio:
Your idea of separating the tests sounds good to me: unit tests should catch any apparent problems with different Python versions, and integration tests should ensure compatibility with different NetBox versions (and it's maybe enough to run the integration tests on one Python version only, "latest-1" or something like that).

markkuleinio:
@raddessi feel free to open another PR for the above (I'm not sure off the bat how to achieve it), I can very well close this one at that point.

arthanson:
superseded by net NetBox 3.3 changes

PR #472, opened 2022-06-14
Fix updating custom filed of type JSON

### LINKED ISSUE #457: Updating object with JSON custom field causes a traceback
I have a JSON custom field on some objects in my NetBox 3.2beta instance. When trying to update one of the objects I get this traceback.

```
  File "/home/jcollie/dev/netbox-maintenance/fix-model-aliases.py", line 602, in <module>
    main()
  File "/usr/lib/python3.10/site-packages/click/core.py", line 1137, in __call__
    return self.main(*args, **kwargs)
  File "/usr/lib/python3.10/site-packages/click/core.py", line 1062, in main
    rv = self.invoke(ctx)
  File "/usr/lib/python3.10/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/lib/python3.10/site-packages/click/core.py", line 763, in invoke
    return __callback(*args, **kwargs)
  File "/home/jcollie/dev/netbox-maintenance/fix-model-aliases.py", line 597, in main
    if device_type.save():
  File "/home/jcollie/.local/lib/python3.10/site-packages/pynetbox/core/response.py", line 529, in save
    updates = self.updates()
  File "/home/jcollie/.local/lib/python3.10/site-packages/pynetbox/core/response.py", line 506, in updates
    diff = self._diff()
  File "/home/jcollie/.local/lib/python3.10/site-packages/pynetbox/core/response.py", line 484, in _diff
    {fmt_dict(k, v) for k, v in self.serialize(init=True).items()}
  File "/home/jcollie/.local/lib/python3.10/site-packages/pynetbox/core/response.py", line 457, in serialize
    ret[i] = flatten_custom(current_val)
  File "/home/jcollie/.local/lib/python3.10/site-packages/pynetbox/core/response.py", line 61, in flatten_custom
    return {
  File "/home/jcollie/.local/lib/python3.10/site-packages/pynetbox/core/response.py", line 62, in <dictcomp>
    k: v if not isinstance(v, dict) else v["value"] for k, v in custom_dict.items()
KeyError: 'value'
```

Kani999:
Close #457 

Since there is no existing PR, I've created one which is based on @markkuleinio solution


FliesLikeABrick:
+1 I installed pynetbox from this branch and it resolved my inability to update JSON custom properties with the error cited in #457

FliesLikeABrick:
@zachmoody  can you review this?  It would be great to have this fix come in upstream to fix our automation and let us avoid having to create an internal fork

abhi1693:
@Kani999  This change does not seem to be the right way to handle this scenario. IMO, you should update the `flatten_custom` to something like

```python
def flatten_custom(custom_dict):
    return {
        k: v if not isinstance(v, dict) else v["id"] for k, v in custom_dict.items()
    }
```

`custom_dict` always contains the `id` field which I believe used to be referred to as `value` in the older versions.

Kani999:
@abhi1693 I've changed the `flatten_custom` method, which basically does nothing at the moment. So it works like a previous change. 

> `custom_dict` always contains the `id` field which I believe used to be referred to as `value` in the older versions.
That's not true. Item contains just a key and value, there is no id or anything else.

rodvand:
Any news on merging this? The issue this PR addresses also affects the ansible modules.

PR #476, opened 2022-06-30
Update doco for all() and filter() methods regarding iteration over result set

sol1james:
Updating doco to reflect the need to encapsulate the result set in a list when calling the all() method of an endpoint.

See disscussion in netbox-community#450 and also sol1james#1


arthanson:
fixed and merged in https://github.com/netbox-community/pynetbox/pull/525

PR #480, opened 2022-07-19
Keep serverdata from base URL

mburggraf:
When querying multi-page data, the current code completely takes the value from the next-parameter in the json answer. 
However if the netbox instance is multi-homed, or accessed unusually, for example by using a tunnel, or using direct IP notation (no working DNS), we may run into an Exception.

With this patch, instead of overriding the complete URL with data from next parameter, we keep the server name/port and scheme with which the api was originally initiated.


abhi1693 (changes requested):
Please provide a test case for this and ideally open a GitHub issue to track if this is still relevant

arthanson:
Closing, not GitHub issue, no tests, no response

PR #488, opened 2022-08-18
Fix for limit without offset being ignored

mrkeathley:
If the limit kwarg is passed without an offset it will functionally be ignored and the entire result set will be returned.

Previously:
```
tenants = netbox_api.tenancy.tenants.all(limit=50)
print(f'Limit Size: {tenants.request.limit}')
print(f'Length of Iterable: {len(tenants)}')
...
Limit Size: 50
Length of Iterable: 5409
```

The simple fix was to provide only the results from netbox when limit or offset were provided. This fixes the output:

```
Limit Size: 50
Length of Iterable: 50
```

### DIFF
--- a/pynetbox/core/query.py
+++ b/pynetbox/core/query.py
@@ -320,8 +320,9 @@ class Request(object):
         req = self._make_call(add_params=add_params)
         if isinstance(req, dict) and req.get("results") is not None:
             self.count = req["count"]
-            if self.offset is not None:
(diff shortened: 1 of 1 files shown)


sebastian-philipp (approved):

arthanson:
I'm actually not seeing a change with this in, still returning all results.  Can you write a test case for this as well.

PR #493, opened 2022-08-28
Fixes path for plugin names with underscores

abhi1693:
For the plugin `netbox-secretstore` and `netbox-access-lists`, I'm unable to use the library because the code auto replaces the `_` with `-`. This causes issues because the path to the APIs require the use of `underscores`


abhi1693:
@zachmoody Ping, let me know if you require more information

PR #507, opened 2022-10-24
Create get_or_raise endpoint

pthmas:
Using the normal get endpoint on Pynetbox doesn't raise any error if no object is returned. This means that we need to constantly check on our side every time get is called to make sure that an object is returned. 
This aims at avoiding: 

```python
if not response:= nb.dcim.locations.get(id=1): 
    raise ValueError("location not found")
```

Using get_or_raise ensure that in the case of no object returned a ValueError is raised.

This way we can also use the python stubs mentioned [here](https://github.com/interdotlink/pynetbox-stubs/blob/main/pynetbox-stubs/_gen/circuits.pyi#L73 ) without having the tag Optional. 

Not that there is also an inconsistency between `get(myid)` and `get(id=myid)`: The former raises a Request error for unknown object, the latter one returns None.

### DIFF
--- a/pynetbox/core/endpoint.py
+++ b/pynetbox/core/endpoint.py
@@ -173,6 +173,39 @@ class Endpoint:
             else:
                 raise e
 
+    def get_or_raise(self, *args, **kwargs):
+        r"""Queries the DetailsView of a given endpoint but raises an error if no object is returned
+        :arg int,optional key: id for the item to be
+            retrieved.
+
+        :arg str,optional \**kwargs: Accepts the same keyword args as
+            filter(). Any search argument the endpoint accepts can
+            be added as a keyword arg.
+
+        :returns: A single :py:class:`.Record` object
+
+        :raises ValueError: if kwarg search returns zero or more than one item
+
+        :Examples:
+
+        Referencing with a kwarg that only returns one value.
+
+        >>> nb.dcim.devices.get_or_raise(name='test1-a3-tor1b')
+        test1-a3-tor1b
+        >>>
+
+        Referencing with an id.
+
+        >>> nb.dcim.devices.get_or_raise(1)
+        test1-edge1
+        >>>
+        """
+
+        resp = self.get(*args, **kwargs)
+        if resp is None:
+            raise ValueError(f"{self.name} not found: {kwargs}")
+        return resp
+
     def filter(self, *args, **kwargs):
         r"""Queries the 'ListView' of a given endpoint.
 


sebastian-philipp:
Any update on this?

nautics889 (changes requested):
I've got nothing against having `get_or_create()` method in the endpoint class.
___
However would you consider to raise custom error instead of `ValueError`? The reason is (as seems to me) that `.get_or_create()` you've implemented calls `.get()` which in it's turn invokes plenty different methods. At least `ValueError` can be raised from [this line](https://github.com/netbox-community/pynetbox/blob/master/pynetbox/core/query.py#L43). I understand we can barely encounter on this cases, but anyway using custom error class can help to differentiate errors to avoid ambiguous.
___
Furthermore, there is `RequestError` in query.py module (at the same directory) which already has [logic to handle 404 NotFound error](https://github.com/netbox-community/pynetbox/blob/master/pynetbox/core/query.py#L43) (at least to persist the context). But `RequestError` accepts a `Request` object in it's `__init__()`. And there is no such object in your `get_or_raise()` method, so you wouldn't raise one from .  
In my opinion original `get()` method should lift up the exception which would be handled by client code. IDK why originally there is [suppressing this exception](https://github.com/netbox-community/pynetbox/blob/master/pynetbox/core/endpoint.py#L183) at all. This is similar to anti-pattern `try: ... except: pass` in fact.
___
My suggestion:
Just create `NotFoundError` at the same module and raise it instead of `ValueError`:
```
class NotFoundError(Exception):
    def __init__(self, name, **kwargs)
        self.message = f"{self.name} not found: {kwargs}"
        super().__init__(self.message)

...

class Endpoint:
    ...

    def get_or_raise(self, *args, **kwargs):
        r"""Queries the DetailsView of a given endpoint but raises an error if no object is returned
        :arg int,optional key: id for the item to be
            retrieved.
        :arg str,optional \**kwargs: Accepts the same keyword args as
            filter(). Any search argument the endpoint accepts can
            be added as a keyword arg.
        :returns: A single :py:class:`.Record` object
        :raises NotFoundError: if kwarg search returns zero or more than one item
        :Examples:
        Referencing with a kwarg that only returns one value.
        >>> nb.dcim.devices.get_or_raise(name='test1-a3-tor1b')
        test1-a3-tor1b
        >>>
        Referencing with an id.
        >>> nb.dcim.devices.get_or_raise(1)
        test1-edge1
        >>>
        """

        resp = self.get(*args, **kwargs)
        if resp is None:
            raise NotFoundError(f"{self.name} not found: {kwargs}")
        return resp

...
```

As for me, it'd be better to remove suppression from [this line](https://github.com/netbox-community/pynetbox/blob/master/pynetbox/core/endpoint.py#L183) at all, but it'd lead to that huge part of scripts using pynetbox face tracebacks with `RequestError` :)

pthmas:
Thank you for your reply!
I mostly agree with you, and will make the changes. I'm pretty busy right now but will find the time later to take care of it.

abhi1693 (changes requested):
Please raise an issue to discuss this change and why this is important

PR #512, opened 2022-11-04
Fixes: Enforce application/json content type

Simkimdm:
With Version 3.3 the Content Type: application/json is enforced by netbox. https://github.com/netbox-community/netbox/commit/bfbf97aec9119539f7f42cf16f52d0ca8203ba60 / https://www.django-rest-framework.org/api-guide/parsers/

It takes long to find the Problem why i am getting a 403 Error just while Post requests.
I finally get it fixed by removing the ";" in all '"Content-Type": "application/json;"'.
Also in all API examples its always called without the ending ";". [docs/integrations/rest-api.md](https://github.com/netbox-community/netbox/blob/21b9732f061b7e91fa77c10da66bd098510354fc/docs/integrations/rest-api.md) / https://github.com/netbox-community/netbox/blob/d2fe59ae8f9c34c1f8a7d51c0d113348d7bed937/netbox/utilities/api.py#L52

Hope this is request is okay?

### DIFF
--- a/pynetbox/core/query.py
+++ b/pynetbox/core/query.py
@@ -163,7 +163,7 @@ class Request:
     def get_openapi(self):
         """Gets the OpenAPI Spec"""
         headers = {
-            "Content-Type": "application/json;",
+            "Content-Type": "application/json",
         }
         req = self.http_session.get(
             "{}docs/?format=openapi".format(self.normalize_url(self.base)),
@@ -185,7 +185,7 @@ class Request:
         present in the headers.
         """
         headers = {
-            "Content-Type": "application/json;",
+            "Content-Type": "application/json",
         }
         req = self.http_session.get(
             self.normalize_url(self.base),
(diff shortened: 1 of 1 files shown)


markkuleinio:
I'm unable to reproduce the 403 error but I'm still in favor of removing the extra semicolons.

```
>>> tag = nb.extras.tags.create(name="TestTag", slug="testtag")
>>> tag
TestTag
>>> pynetbox.__version__
'6.6.2'
>>> nb.status()["netbox-version"]
'3.3.7'
>>>
```
access.log:

```
172.16.10.10 - - [07/Nov/2022:11:00:34 +0200] "POST /api/extras/tags/ HTTP/1.1" 201 251 "-" "python-requests/2.28.0"
```

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type says that the semicolon is used for separating the charset.

nautics889 (changes requested):
Looks good to me, but there's one more excessive `;` in [test_request.py module](https://github.com/netbox-community/pynetbox/blob/master/tests/unit/test_request.py#L13).

Could you please remove this also?

abhi1693:
This will be superseded by #547 when its merged.

PR #521, opened 2022-12-13
Remove pre-NetBox 3.3 features

markkuleinio:
Removes references to `secrets` app (removed in NetBox 3.0) and the related private key and session key references since `pynetbox` 7.x is now for NetBox 3.3(+) only.

Also removes
- `App.choices()` (only worked with NetBox 2.7.x and older)
- `App.custom_choices()` (only worked with NetBox 2.9.x and older)


arthanson (approved):

arthanson:
Remove old NetBox < 3.3 code.

PR #533, opened 2023-02-15
Fix __str__ on Record to always return a string

Frazew:
# Summary

On certain edge cases, it is possible for the `__str__` method of a Record object to return something that is not of type `str`, which raises an Exception: `TypeError: __str__ returned non-string (type Record)`

# Steps to reproduce

```python
import pynetbox
nb = pynetbox.api("http://localhost:8080", token="0123456789abcdef0123456789abcdef01234567")

# create test platform
platform = nb.dcim.platforms.create(
    name="test platform",
    slug="test-platform",
    napalm_args={
        "anything": {
            "name": {
                "attr1": "value1",
                "attr2": "value2",
            }
        },
    }
)
```

This is one of the ways this bug will be triggered. It can happen on any object can hold arbitrary JSON data in a field that is not explicitly mapped to a `JsonField` class in pynetbox (for example a `config_context` on `dcim.devices` will not trigger the bug)

# Description

When a generic Record is generated from data returned by the API, each field is dynamically mapped deeply to find candidate Records with other well-known objects types. However, some netbox objects can hold arbitrary JSON, and if that JSON (dictionary) is not explicitly mapped to a `JsonField`, it will be interpreted as a Record object. Printing the Record will hence look for the keys that should represent it (`name`, `label`, `display`) and otherwise return an empty string. Given that these fields can be set to arbitrary values, the data can be of any type, which breaks the implicit assumption that it should be a `str`.

# Fix

The PR simply add a call to `str` on the result that should be returned.

### DIFF
--- a/pynetbox/core/response.py
+++ b/pynetbox/core/response.py
@@ -323,7 +323,7 @@ class Record:
         return dict(self)[k]
 
     def __str__(self):
-        return (
+        return str(
             getattr(self, "name", None)
             or getattr(self, "label", None)
             or getattr(self, "display", None)


nautics889 (changes requested):
I wouldn't check if `value` is a string comparing `type(value)` to `str` directly. Instead just use
```
if not isinstance(value, str):
    value = ""
```
This is because `isinstance()` function [consider inherited classes accroding to Liskov Susbs. principle](https://stackoverflow.com/questions/1549801/what-are-the-differences-between-type-and-isinstance) unlike direct type comparision:
```
>>> class Bar: 
...     pass
>>> class Foo(Bar):  # child class
...     pass
>>> f = Foo()
>>> isinstance(f, Bar)
True
>>> type(f) == Bar
False  # that is the point
```

Also, could we just let it be something like:
```
    def __str__(self):
        # just set empty string as a default for __getattr__()
        value = str(getattr(self, "name", "")
                    or getattr(self, "label", "")
                    or getattr(self, "display", ""))
        return value
```
?
I've tested the code above, it works correct.

I'd appreciate it if you updated your PR considering the points i described. 
This PR fixes a real import bug and should be merged.

Frazew:
You're absolutely right about the `type` check, I naively thought that `__str__` would not accept returning `str` subclasses but it absolutely does, a `isinstance` check would've worked better.

Anyway your suggestion is a lot better while providing the same functionality, I updated the PR accordingly. I added the last `or ""` case: it might happen that `getattr(self, "display", "")` returns `None`, which would leaf to return the string `"None"` instead of an empty string as is currently the case.

Thank you!

markkuleinio (commented):
Note that changing `None`s to empty strings does not actually change the behaviour: the last part of the `or` chain (the empty string) will be returned anyway if none of the attributes `name`, `label` or `display` exists.

For me it's standard to use `None` in `getattr()` default value when explicitly catching the "not exist" cases.

Frazew:
Updated with `None` since indeed it's the same thing. The PR is down to a super simple addition now.

---

Note: the initial reason for type-checking `str` before returning was for cases like this:
```python
platform = nb.dcim.platforms.create(
    name="test platform",
    slug="test-platform",
    napalm_args={
        "name": {
            "display": ["test"]
        },
    }
)
print(platform.napalm_args)
```
which outputs the string `"['test']"`.

This is also obviously true with other types (e.g. `"name": 42` or `"name": 4.3`). I think it boils down to personal preference: returning a string representation for a number is not weird per se, but that means also returning a string representation for lists, which feels weird given that the expectation for `str` on a `Record` object is to have a user-friendly name. What do you think?

nautics889 (approved):
I like the current implementation!

___

As for:
>that means also returning a string representation for lists, which feels weird given that the expectation for str on a Record object is to have a user-friendly name

Well, as you've mentioned, if we go back to the initial option (with type-checking), like:
```
if not isinstance(value, str):
    return ""
```
we will get an empty string when the object has list-type `name`/`label`/`display` attribute like `["test_name"]`. I guess it'd be better to display user a non-user-friendly list representation rather than empty string. 

So i'd prefer to keep the current option, you've uploaded.

Thanks for the update!

abhi1693:
I think a better way would be to map the field with a type rather than forcing the function to always return string. This use case is very limited to the change at hand and I'd discourage it until it can be tested more thoroughly for all or most other use cases.

Moreover, if this is still an issue that requires to be solved, please open up an issue and discuss over there.

PR #545, opened 2023-04-24
replace None with 'null' in query parameters

doc-sheet:
Requests skips parameters that are None.
Map 'null' string to None python type for a more native coding

eg: `nb.dcim.interfaces.filter(mac_address__n=None)`


abhi1693 (changes requested):
Could you provide a test case for this?

abhi1693 (approved):

PR #547, opened 2023-04-30
Update OpenAPI endpoint for 3.5

devon-mar:
This breaks compatibility for older versions of NetBox.

If compatibility is desired, we can try the v3.5 endpoint first then attempt the older endpoint.

### DIFF
--- a/pynetbox/core/query.py
+++ b/pynetbox/core/query.py
@@ -153,10 +153,10 @@ class Request:
     def get_openapi(self):
         """Gets the OpenAPI Spec"""
         headers = {
-            "Content-Type": "application/json;",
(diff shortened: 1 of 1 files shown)


K0HAX:
I think the schema URL needs to be `{}schema/?format=json`, rather than `{}schema` for this to work. At least in my testing that's what I had to do.

devon-mar:
@K0HAX I misread the header that we're sending as `Accept` for some reason, that's why I left it out. I'm not sure why the original implementation was sending the `Content-Type` header in the request so I changed that to `Accept` so that NetBox returns JSON.

abhi1693:
This will be superseded by #564 once that's merged.

PR #554, opened 2023-06-14
Get faster response for count() method

srfwx:
By using `brief=1` parameter in the get_count() method, we can reliably get a much faster response.

As an example, without the parameter, running 100 iterations of these count action to my cloud NetBox instance:
nb.dcim.cables.count() : average duration 0.1 sec 
nb.ipam.ip_addresses.count() : average duration 0.47 sec

By adding `brief=1` :
nb.dcim.cables.count() : average duration 0.075 sec
nb.ipam.ip_addresses.count() : average duration 0.343 sec

### DIFF
--- a/pynetbox/core/query.py
+++ b/pynetbox/core/query.py
@@ -409,5 +409,5 @@ class Request:
         """
 
         if not hasattr(self, "count"):
-            self.count = self._make_call(add_params={"limit": 1})["count"]
+            self.count = self._make_call(add_params={"limit": 1, "brief": 1})["count"]
         return self.count


pynetbox/core/query.py
```diff
@@ -409,5 +409,5 @@ def get_count(self, *args, **kwargs):
         """
 
         if not hasattr(self, "count"):
-            self.count = self._make_call(add_params={"limit": 1})["count"]
+            self.count = self._make_call(add_params={"limit": 1, "brief": 1})["count"]
```
abhi1693, line 412: This is just a minor nitpick, but I would prefer this to be set to `True`. Although, python evaluates `1` the same but a true boolean value conveys the meaning and intention much better.

abhi1693 (changes requested):

pynetbox/core/query.py
srfwx: Hi,
NetBox documentation specifically refers to brief=1
https://demo.netbox.dev/static/docs/rest-api/overview/

abhi1693 (approved):

Issue #548, opened 2023-05-01
No longer works with Netbox as of v3.5.0 due to OpenAPI change


K0HAX:
When targeting a Netbox instance of v3.5.0 I get the error `pynetbox.core.query.RequestError: The requested url: https://<url>/api/docs/?format=openapi could not be found.` when calling `pynetbox.api('<url>', token='<token>')`

markkuleinio:
Creating the API object works here:

```
>>> import pynetbox
>>> pynetbox.__version__
'7.0.1'
>>> netbox = pynetbox.api("http://netbox-test.lein.io/", token="176d4c04ccc8f2a549ea6fd39312367d9da5a796ff")
>>> netbox.version
'3.5'
>>> netbox.status()["netbox-version"]
'3.5.0'
>>>
```

I see you already noticed https://github.com/netbox-community/pynetbox/pull/547 that deals with the OpenAPI change.

photonicnet:
Is the change in #547 and the commit going into 7.0.2?  We have not noticed anything else with netbox 3.5.x other than the OpenAPI change at this stage.  We have not worked with any of the new features as of yet though.

PR #564, opened 2023-08-24
Adds support for NetBox v3.5+

abhi1693:
Fixes: #548


jeremystretch (approved):
Excellent work @abhi1693! Thanks for taking the time to add v3.5 support.

arthanson (approved):
Great work @abhi1693 !

PR #576, opened 2023-08-30
Adds support for netbox v3.6

abhi1693:


tests/integration/test_dcim.py
```diff
@@ -188,13 +197,21 @@ def init(self, request, interface):
...
-    def power_outlet(self, api, device_type, device_role, site):
-        pdu = api.dcim.devices.create(
-            name="test-pdu",
-            device_role=device_role.id,
-            device_type=device_type.id,
-            site=site.id,
-        )
+    def power_outlet(self, api, device_type, role, site):
+        if version.parse(api.version) >= version.parse("3.6"):
+            pdu = api.dcim.devices.create(
```
arthanson, line 202: Why are there two of these, they are the exact same?  If one is for the older version shouldn't it be device_role, also does the old version test even make sense when the fixtures have changed to just device

arthanson (changes requested):

arthanson (approved):

PR #577, opened 2023-09-02
Fix device_role to role breaking change in Netbox 3.6.0

shaul75:


abhi1693:
This is incompatible with the test cases for 3.3 to 3.5. Moreover, if you wish to submit a PR please discuss with me on the issue first to avoid submitting code that is already being worked on.

shaul75:
> This is incompatible with the test cases for 3.3 to 3.5. Moreover, if you wish to submit a PR please discuss with me on the issue first to avoid submitting code that is already being worked on.

Thanks @abhi1693 
I expected this was already being worked on (or planned for). The reason I didn't discuss it first is that I already made it for my local environment anyway, so I submitted this PR.

Issue #591, opened 2023-11-03
Choices are not returned when the API token can PUT but not POST on an endpoint


BenoitKnecht:
### pynetbox version

v7.1.0

### NetBox version

v3.4.3

### Python version

3.9

### Steps to Reproduce

Using an API token that cannot POST on the `/dcim/devices/` endpoint for instance, try getting choices using

```python
nb.dcim.devices.choices()
```

### Expected Behavior

A list of choices is returned.

### Observed Behavior

```
ValueError: Unexpected format in the OPTIONS response at https://netbox.example.com/api/dcim/devices
```

BenoitKnecht:
I can submit a PR for a simple fix that would get the choices from the `.actions.PUT` sections returned by the `OPTIONS` request when `.actions.POST` is missing.

markkuleinio:
Just wondering: what is the use case for fetching the choices list when there is no possibility anyway to use the choice values to create/edit objects?

Or do I miss something about POST vs PUT

BenoitKnecht:
> Just wondering: what is the use case for fetching the choices list when there is no possibility anyway to use the choice values to create/edit objects?
> 
> Or do I miss something about POST vs PUT

@markkuleinio With POST you create new objects, and with PUT you update existing objects. So fetching choices is useful in that latter case too. :slightly_smiling_face:

markkuleinio:
Right, so the use case is a user that doesn't have create access but has modify access.

BenoitKnecht:
Hey, could someone review (and ideally merge) the PR above (#592)? I'm still hitting this bug whenever I use Ansible to interact with Netbox.

PR #592, opened 2023-11-05
core/endpoint: Get choices from PUT if POST missing

BenoitKnecht:
### Fixes: #591

If the API token is not allowed to POST on a particular endpoint, choices won't be returned in `.actions.POST`, but we can still get them in `.actions.PUT`.

This commit falls back to getting choices from `.actions.PUT` instead of raising an exception when `.actions.POST` is missing.

### DIFF
--- a/pynetbox/core/endpoint.py
+++ b/pynetbox/core/endpoint.py
@@ -587,9 +587,10 @@ class Endpoint:
             token=self.api.token,
             http_session=self.api.http_session,
         ).options()
-        try:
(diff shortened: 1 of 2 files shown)


arthanson:
@BenoitKnecht Can you please add a test case for this?

BenoitKnecht:
@arthanson Could you take another look at this PR? I'd really love to get this merged.

arthanson (approved):

Issue #597, opened 2023-12-12
pynetbox greater than 7.0.0 fails to insert complex custom_fields


costasd:
### pynetbox version

v7.2.0

### NetBox version

v3.5.9

### Python version

3.11

### Steps to Reproduce

Hello,

in our setup we introduced a complex custom_field in each device that is of the following form, in python notation:
```python
{
  "metadata_config" : [
    {"a" : "b", "c": "d"},
    {"a":  "e", "c": "k"}
  ]
}
```
I found out that any version over 7.0.0, essentially anything containing the #518 fix that rewrote `flatten_custom`, is breaking updating this custom_field via pynetbox. `update()` returns fine, but all dicts are set as `None` into netbox.

The issue seems to lie with a cornercase in `flatten_custom`. 
Below you can find a way to test this with 7.0.0, 7.2.0 and a candidate fix that I'll open a PR with.

Thanks!

to run it, just execute `./run.sh`, after writing adding these two files into some directory.

__filename: run.sh__
```bash
#!/bin/bash

pip install pytest

echo "v7.0.0 works"
pip install pynetbox==7.0.0
pytest

echo "v7.2.0 fails"
pip install pynetbox==7.2.0
pytest

echo "candidate fix works"
pip install git+https://github.com/costasd/pynetbox@6fe8494f74cbe7779c98cb949af2ee8d95e74d0e
pytest
```
__filename: test_custom_fields.py__
```python
import pynetbox

class TestComplexCustomFields:
    def test_delete_custom_field(self):
        nb = pynetbox.api('http://localhost:8080', token='0123456789abcdef0123456789abcdef01234567')
        field = nb.extras.custom_fields.get(name="metadata_config")
        nb.extras.custom_fields.delete([field])

    def test_create_custom_field(self):
        nb = pynetbox.api('http://localhost:8080', token='0123456789abcdef0123456789abcdef01234567')
        field = nb.extras.custom_fields.create(
            name="metadata_config",
            content_types=["dcim.device"],
            type="json",
            description="Format: [{\"key1\": \"value1\", \"key2\": \"value2\"}]",
            data_type="object",
            default=[])
        assert field

    def test_complex_structures_to_custom_field(self):
        nb = pynetbox.api('http://localhost:8080', token='0123456789abcdef0123456789abcdef01234567')

        device = nb.dcim.devices.get(1)
        dev_id = device.id

        # A list containing a dict
        data = [
            [],
            [{"a": "b"}],
[...]

PR #598, opened 2023-12-12
flatten_custom: return dict if id is not present

costasd:
### Fixes: #597 

If we find a free form dict in the value, containing no id, let's return it as it is.


oyacky:
Merge please

arthanson:
@costasd can you please add a test for this?

Issue #600, opened 2024-01-03
Add classifier and testing for Python 3.12


markkuleinio:
### pynetbox version

v7.3.0

### NetBox version

v3.7.0

### Feature type

Change to existing functionality

### Proposed functionality

Add Python 3.12 in setup.py classifiers and in the test runs.

### Use case

Python 3.12 has been out for three months, and the first fix release 3.12.1 is also out already, so let's test against it as well.

abhi1693:
@markkuleinio Would you like to submit a PR?

markkuleinio:
That'll be tomorrow (EET) but yes

arthanson:
Closed, update to support Python 3.12

PR #601, opened 2024-01-04
Add Python 3.12 in tests and other lists

markkuleinio:
### Fixes: #600

Added Python 3.12 in tests, setup.py classifiers and in the bug report forms.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yaml
@@ -31,7 +31,6 @@ body:
         - "3.10"
         - "3.11"
         - "3.12"
-        - "3.13alpha/beta"
     validations:
       required: true
   - type: textarea
--- a/.github/ISSUE_TEMPLATE/bug_report.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yaml
@@ -11,14 +11,14 @@ body:
     attributes:
       label: pynetbox version
       description: What version of pynetbox are you currently running?
-      placeholder: v7.1.0
+      placeholder: v7.3.0
     validations:
       required: true
   - type: input
     attributes:
       label: NetBox version
       description: What version of NetBox are you currently running?
-      placeholder: v3.6.0
+      placeholder: v3.7.0
     validations:
       required: true
   - type: dropdown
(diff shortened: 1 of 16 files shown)


markkuleinio:
I did not change `.readthedocs.yaml` yet, I guest that's just for building the docs, maybe that can be done later.

markkuleinio:
I don't know how to fix the failing tests, @abhi1693 

>  Container pytest_pynetbox_1704357384-netbox_v3_3_netbox-housekeeping-1  Created
Error response from daemon: failed to mkdir /var/lib/docker/volumes/pytest_pynetbox_1704357384_pytest_pynetbox_v3_3_netbox-media-files/_data/image-attachments: mkdir /var/lib/docker/volumes/pytest_pynetbox_1704357384_pytest_pynetbox_v3_3_netbox-media-files/_data/image-attachments: file exists

abhi1693:
It's possible that 3.12 was not supported with older netbox versions.

markkuleinio:
> It's possible that 3.12 was not supported with older netbox versions.

Why would that be shown as a docker volume creation error? (Sorry, I'm not a dockerist)

markkuleinio:
I'll still try with NetBox 3.7 only, just a minute

markkuleinio:
I don't see the logic: how it failed the Python 3.8 + NetBox 3.3 test. Something is wrong in the docker logic, cannot help that. Restoring the original changes now

markkuleinio:
Looks like the docker image for NetBox 3.3 is not working, others are fine.

.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -30,6 +30,8 @@ body:
         - "3.9"
         - "3.10"
         - "3.11"
+        - "3.12"
+        - "3.13alpha/beta"
```
abhi1693, line 34: I don't think we should test on 3.13 until it's stable release

abhi1693 (changes requested):

.github/ISSUE_TEMPLATE/bug_report.yaml
markkuleinio: It's not the test list, it's the bug report template dropdown menu, making it possible for submitting a bug report when running any currently available Python version:

image

abhi1693: We will add it when we are running tests for it. I don't want to give an illusion to the end user that this version is also supported

abhi1693 (approved):

PR #608, opened 2024-03-11
Relax upper limit of packaging version

frenzymadness:
It seems to me that pynetbox should work with packaging 24 but I'm not able to run all tests because I don't have docker installed.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,2 +1,2 @@
 requests>=2.20.0,<3.0
-packaging<24.0
+packaging
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,7 @@ setup(
     long_description_content_type='text/markdown',
     install_requires=[
         "requests>=2.20.0,<3.0",
-        "packaging<24.0"
+        "packaging"
     ],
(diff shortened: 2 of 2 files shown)


frenzymadness:
It builds fine in Fedora Linux with packaging 24.0

markkuleinio:
If I looked it correctly, `packaging` is only used for `version`. To be honest, I find it to be such a simple use and core function of `packaging` that I don't think it needs any kind of version number pinning.

frenzymadness:
> If I looked it correctly, `packaging` is only used for `version`. To be honest, I find it to be such a simple use and core function of `packaging` that I don't think it needs any kind of version number pinning.

That sounds even better from my point of view. Should I update this PR?

markkuleinio:
I'd say so yes.

frenzymadness:
Done and force-pushed.

gaige:
I'd appreciate seeing this one accepted.

I am not sure if others are encountering this, but the <24.0 requirement is causing some difficulties in aligning with code that I'm using that wants packaging 24.x.

PR #613, opened 2024-05-07
Include Auth Token when Getting API Version

terraflubb:
### Fixes: #612

When we get the API version, we aren't including the auth token if we have one, and as a result the request fails (as of Netbox 4.0.0).

I simply copied the header-building from `get_status` below to make sure I was following the convention, since it does exactly what we'd like.

I tested this locally and it resolved my issue, but I haven't tested it fully throughout our API integrations (as I imagine there will be more as we move to Netbox 4.0.0) so I'm not sure if there are any other side effects I haven't considered.

PR #616, opened 2024-05-27
Fix API version for Netbox version > 4.x

### LINKED ISSUE #612: Auth fails getting API version with Netbox 4.0.0
### pynetbox version

v7.3.3

### NetBox version

v4.0.0

### Python version

3.11

### Steps to Reproduce

1. Upgrade to Netbox 4.0.0
2. Run this (auth is correct in the environment)

```python
import pynetbox
import os

NETBOX_URL = os.environ.get('NETBOX_URI')
NETBOX_KEY = os.environ.get('NETBOX_API_KEY')

nb = pynetbox.api(url=NETBOX_URL, token=NETBOX_KEY)

print(nb.version)
```

### Expected Behavior

It should say:

```
'4.0'
```

### Observed Behavior

This will throw:

```
pynetbox.core.query.RequestError: The request failed with code 403 Forbidden: {'detail': 'Authentication credentials were not provided.'}
```

### Fix

The issue is that in `get_version`, we aren't including any auth in the header, which was never a problem until Netbox 4.0.0. One could argue either way if it's an issue with Netbox or `pynetbox`, but the fix is pretty harmless:

By turning:
https://github.com/netbox-community/pynetbox/blob/master/pynetbox/core/query.py#L188-L190

Into something more like:
https://github.com/netbox-community/pynetbox/blob/master/pynetbox/core/query.py#L206-L208

This issue stops! :tada: 

~~I have a PR to fix this which I can't create until this issue exists.~~ Suggested fix is here #613

btriller:
fixes #612

### DIFF
--- a/pynetbox/core/query.py
+++ b/pynetbox/core/query.py
@@ -192,7 +192,7 @@ class Request:
             self.normalize_url(self.base),
             headers=headers,
         )
-        if req.ok:
+        if req.ok or req.status_code == 403:
             return req.headers.get("API-Version", "")
         else:
             raise RequestError(req)


terraflubb:
I think this is a nicer solution to #612 than mine that captures more potential cases (#613 won't solve the "I don't have auth, just curious what version, which isn't a secret anyway"), plus it's smaller. But I'll take whatever I can get at this point.

But FYI @btriller , the workflow that does the linting (which requires approval and who knows when that will happen :cry:) will fail as I discovered. It looks like this check was added without first being executed on the code base.

If you cherry-pick this commit from my PR: https://github.com/netbox-community/pynetbox/pull/613/commits/603601696e06237898b5f6d6f49ba5d8737a5862

It will add all the newlines needed to pass the linting.

Speaking from experience, it would be disappointing if this finally got its workflow approved only to be stalled by another month for reasons outside your control.

jeffgdotorg (approved):
It doesn't get much cleaner or more narrowly targeted than this. Thanks.

terraflubb:
:heart: Thank you so much for jumping in @jeffgdotorg !

PR #637, opened 2024-08-12
Add some tests for core.response.flatten_custom

costasd:
Relates to #597

### Fixes: #1234

This changeset adds a set a test for `flatten_custom`.


costasd:
@arthanson this contains a basic test for `flatten_custom` fixes that were introduced in #598.

PR #641, opened 2024-08-30
API token is passed to retrieve NetBox version

### LINKED ISSUE #640: `nb.version` Returns an Empty String Due to OIDC Proxy Configuration
### pynetbox version

v7.4.0

### NetBox version

v4.0.9

### Python version

3.11

### Steps to Reproduce

**Description:**

When using NetBox with an OIDC (OpenID Connect) proxy for authentication, the `nb.version` property consistently returns an empty string (`''`). This issue arises because the `get_version` method in `pynetbox` does not receive the `API-Version` header, as the request is redirected to the OIDC login page due to missing authorization.

**Steps to Reproduce:**

1. Set up NetBox behind an OIDC proxy with the following Apache configuration:

   ```apache
   <Location /api>
       <If "-z req('authorization')">
         Require valid-user
       </If>
       <Else>
          Require ip <IP_ADDRESS_RANGE>
       </Else>
   </Location>
   ```

2. Use pynetbox to access the API.

3. Call the `nb.version` property.

4. Observe that `nb.version` returns an empty string.

**Expected Behavior:**

The `nb.version` property should return the correct NetBox version string.

**Actual Behavior:**

The `nb.version` property returns an empty string.

**Cause:**

The OIDC proxy redirects unauthorized requests to the login page, causing the `get_version` method to miss the `API-Version` header.

**Proposed Solution:**

Modify the `version` property in `pynetbox/core/api.py` to ensure the authorization token is passed to the `get_version` method. Here's the proposed change:

```python
@property
def version(self):
    """Gets the API version of NetBox.

    Can be used to check the NetBox API version if there are
    version-dependent features or syntaxes in the API.

    :Returns: Version number as a string.
    :Example:

    >>> import pynetbox
    >>> nb = pynetbox.api(
    ...     'http://localhost:8000',
    ...     token='your_token_here'
    ... )
    >>> nb.version
    '3.1'
    >>>
    """
    version = Request(
        base=self.base_url,
        token=self.token,  # Ensure the token is passed here
        http_session=self.http_session,
    ).get_version()
    return version
```

**Additional Context:**

The issue is specific to environments where NetBox is protected by an OIDC proxy, requiring a valid token for API access.

### Expected Behavior

> nb.version
'4.0'

### Observed Behavior

> nb.version
''

Kani999:
### Fixes: #640 

- Updated the `version` property in `pynetbox/core/api.py` to pass the API token when calling `get_version`.
- No problem raised even if the token is None
- This change addresses the issue where `nb.version` returns an empty string when using an OIDC proxy, due to unauthorized requests being redirected to the login page.

Closes #640

### DIFF
--- a/pynetbox/core/api.py
+++ b/pynetbox/core/api.py
@@ -110,6 +110,7 @@ class Api:
         """
         version = Request(
             base=self.base_url,
+            token=self.token,
             http_session=self.http_session,
         ).get_version()
         return version


rondemena (approved):
this is a non-breaking unfortunately necessary change to get operations successful without local circumventions, can we get this drawn in?

rondemena:
It was not my intent to trigger git checks and move to the next step but to bring relevance to the fix is desired in other places and reinforce the proposed change, apologies to the maintainers.

Kani999:
> someone correct me if I'm wrong but isn't this already addressed and merged in #616 I've stumbled over the same problem regarding [netbox-community/Device-Type-Library-Import#134](https://github.com/netbox-community/Device-Type-Library-Import/issues/134)

I don't think so. With our specific configuration, where a redirect occurs when the user is not authorized, it won't trigger a 403 error.

TonkSlip:
I have the same issue, and I need this fix too

manuelrais:
Hi  
Same here, any idea when this request will be merged ? :)

abhi1693:
@arthanson Can you check this and if possible merge it?

arthanson (approved):

Issue #658, opened 2024-12-04
Speed improvements on Record init


srfwx:
### pynetbox version

v7.4.1

### NetBox version

v4.1.7

### Feature type

Change to existing functionality

### Proposed functionality

Hi,

We've built a caching layer to Pynetbox where responses are stored as json inside a Redis cache, then Record objects are reconstructed from there on cache hit. 

However I've noticed that the `_init_` of a `Record` object can still take a substantial amount of time due to the various processing steps

As a first step I have:
- simplified the `_endpoint_from_url` method so that it doesn't need `urlsplit` leading to a 50% speed improvement of that method
- used marshall dump/load instead of deep_copy for cache initialization
- refactored the `_parse_values` method to limit the number of instance type checking and number of key/value that need further processing

All in all, according to a few benchmarks I made, the whole Record init shows more than 50% speed improvement. Below is an example when loading 14407 records.

Benchmark using cProfile with Pynetbox v7.4.1
```
      ncalls  tottime  percall  cumtime  percall filename:lineno(function)
142153/14407    0.829    0.000    8.681    0.001 .venv/lib/python3.10/site-packages/pynetbox/core/response.py:278(__init__)
```
Benchmark using cProfile with our fork
```
      ncalls  tottime  percall  cumtime  percall filename:lineno(function)
142153/14407    0.343    0.000    4.037    0.000 .venv/lib/python3.10/site-packages/pynetbox/core/response.py:263(__init__)
```

Would you be interested in having any/all of these things merged into upstream?

### Use case

General speed improvement of the library

### External dependencies

None

srfwx:
Latest benchmark after adding a cache of endpoints on `App` object and cache of apps on `PluginsApp`, reaching now a 70% speed improvement.
```
      ncalls  tottime  percall  cumtime  percall filename:lineno(function)
142153/14407    0.220    0.000    2.763    0.000  .venv/lib/python3.10/site-packages/pynetbox/core/response.py:263(__init__)
```

arthanson:
@srfwx can you provide a PR for it so we can review it?  It sounds beneficial but hard to say without seeing the exact code.

srfwx:
Hi,
Yes of course, here is the PR #659 
I followed the contribution guidelines and opened an issue first.
The changes seem to work as expected in real-world usage, but some tests are currently failing. I'm open to feedback or guidance on how to address these.
Thank you for your help!

srfwx:
[...]

Issue #625, opened 2024-06-26
List of dictionaries as a field causes Record.save() error


MaayanPeleg:
### pynetbox version

v7.3.3

### NetBox version

V3.7.8

### Python version

3.9

### Steps to Reproduce

In a plugin model, create a JSON field that is set to a lost of dictionaries.
For example: [{"example": "bug"}]

### Expected Behavior

This field should be handles as a JSON list

### Observed Behavior

Pynetbox on initialising the record from netbox, treats the list of dictionaries as references to objects in netbox, so each dictionary is initialised as records, when save is ran, it trys to get it's records id and raises an error as there is no id, as it's not a Record. The bug appears to be caused in _init_cache aswell as in _parse_values() in list_parser()

cis-ks:
I can confirm that the problem also occurs with plugins that have JSON fields. Unfortunately, in pynetbox, the field type is not changed to JSONField for plugin models like it is for core models, which are defined separately.

dxks:
I can confirm that the problem also occurs with plugins that have JSON fields. Unfortunately, in pynetbox, the field type is not changed to JSONField for plugin models like it is for core models, which are defined separately.

Kani999:
I believe that this issue has been addressed in the latest release, `pynetbox 7.4.0`. You can find more information about this release [here](https://github.com/netbox-community/pynetbox/releases/tag/v7.4.0).

This update resolves the following related issues and pull requests:
- Issue: [#597](https://github.com/netbox-community/pynetbox/issues/597)
- Pull Request: [#598](https://github.com/netbox-community/pynetbox/pull/598)

PR #659, opened 2024-12-04
Draft: Speed improvements

srfwx:
### Fixes: #658 #625

Various refactoring allows for a substantial increase in performance of the Record initialization.

- `_parse_values` was heavily refactored to avoid multiple similar checks and unnecessary processing
- in turn, `get_return` now only needs to work on `Record` object and drops legacy code from Netbox 2.7
- `_endpoint_from_url` can work with string instead of using the `urlsplit` library
- finally caching of apps/endpoints avoid constant reinstantiation of similar objects

### DIFF
--- a/pynetbox/core/endpoint.py
+++ b/pynetbox/core/endpoint.py
@@ -637,8 +637,8 @@ class Endpoint:
 
         if any(i in RESERVED_KWARGS for i in kwargs):
             raise ValueError(
-                "A reserved {} kwarg was passed. Please remove it " 
-                "try again.".format(RESERVED_KWARGS)
+                "A reserved kwarg was passed ({}). Please remove it "
+                "and try again.".format(RESERVED_KWARGS)
             )
 
         ret = Request(
--- a/pynetbox/core/endpoint.py
+++ b/pynetbox/core/endpoint.py
@@ -637,9 +637,8 @@ class Endpoint:
 
         if any(i in RESERVED_KWARGS for i in kwargs):
             raise ValueError(
-                "A reserved {} kwarg was passed. Please remove it " "try again.".format(
-                    RESERVED_KWARGS
-                )
+                "A reserved {} kwarg was passed. Please remove it " 
+                "try again.".format(RESERVED_KWARGS)
             )
 
         ret = Request(
--- a/pynetbox/core/endpoint.py
+++ b/pynetbox/core/endpoint.py
@@ -53,6 +53,7 @@ class Endpoint:
             endpoint=self.name,
         )
         self._choices = None
+        self._cache = None
 
     def _lookup_ret_obj(self, name, model):
         """Loads unique Response objects.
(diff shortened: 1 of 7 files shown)


srfwx:
@arthanson the size of this PR may have gotten a little bit out of hand.
I assume you may want to break it into into smaller changes.

As a summary here is what's been implemented so far:

- Cache of Endpoints on App
- Cache nested Records on Endpoints
- Adds class ValueRecord for dict that don't include an endpoint (contrary to class Record)
- Improves speed of computation of endpoint from URL
- Lazy computation of Record.endpoint
- Improves value parsing performance and readability

Please let me know what you think and what you'd like to possibly see as individual PR.

pynetbox/models/mapper.py
```diff
@@ -109,3 +109,5 @@
     "wireless.WirelessLANGroup": None,
     "wireless.wirelesslink": None,
 }
+
+TYPE_CONTENT_MAPPER = {v: k for k, v in CONTENT_TYPE_MAPPER.items() if v is not None}
```
arthanson, line 113: What is this for? I don't see it used anywhere?

pynetbox/models/dcim.py
```diff
@@ -151,6 +151,7 @@ def __str__(self):
 
 
 class Interfaces(TraceableRecord):
+    device = Devices
```
arthanson, line 154: Can you put this in a separate PR, it doesn't look like a speed improvement but a bug fix (if a bug fix and there isn't an issue for it, please open an issue).

pynetbox/core/endpoint.py
```diff
@@ -53,8 +84,12 @@ def __init__(self, api, app, name, model=None):
             endpoint=self.name,
         )
         self._choices = None
+        self._init_cache()
```
arthanson, line 87: This is only called in one place and has a function defined for it - can you just put it inline with a comment instead, not sure why we need a separate function for this?

pynetbox/core/endpoint.py
```diff
@@ -20,6 +20,37 @@
 RESERVED_KWARGS = ()
 
 
+class CachedRecordRegistry:
```
arthanson, line 23: Maybe? Split this and the one in core/app.py into two separate PR's?  Although both are caching - the seem like functionally separate functionality that can be reviewed separately (and potentially backed out separately if any issues).

arthanson (changes requested):

jnovinger:
@srfwx , can you address/answer Arthur's concerns/questions? Otherwise, we'll need to close this out as stale work before too long.

srfwx:
Hi @jnovinger ,
Yes sure, I'll get to it as soon as possible.
Note that it took 5 months to get initial feedback so as you can imagine I've switched to other topics in the meantime... But I'll try to allocate time for this in the coming couple of weeks if that's alright

arthanson:
Thank for getting back. I'll go ahead and close this for now @srfwx please reopen when you are ready to get back to it.

Issue #674, opened 2025-04-11
Update uri_to_obj_class_map to support additional supported termination endpoints in trace() method


bcollet:
### pynetbox version

v7.4.1

### NetBox version

v4.2.3

### Feature type

Data model extension

### Proposed functionality

When tracing a circuit, only the following hops type will be mapped to the corresponding class:

- `Cables`
- `FrontPorts`
- `Interfaces`
- `RearPorts`

I think it would be beneficial for at least the following types to be included:
- `CircuitTerminations`
- `ConsolePorts`
- `ConsoleServerPorts`
- `PowerOutlets`
- `PowerPorts`

Without it, there is no easy nor standard way to identify the type of the object along the traced path.

### Use case

When trying to programmatically determine the type of a hop along a traced path, having all types mapped to their corresponding classes allow the use of `isinstance()`.

PR #675, opened 2025-04-11
Update uri_to_obj_class_map to support additional supported termination endpoints

bcollet:
### Closes: #674 

The PR proposes to add the following types to `uri_to_obj_class_map` to return the appropriate object type when tracing a path:

- CircuitTerminations
- ConsolePorts
- ConsoleServerPorts
- PowerOutlets
- PowerPorts


arthanson (changes requested):
@bcollet sorry for the late response on this, getting caught up after NetBox 4.3 work.  This looks good and thank you for the work. I think it also needs PowerFeed (looking at https://github.com/netbox-community/netbox/blob/main/netbox/dcim/constants.py#L99) Can you please check and add that one is as well.

bcollet:
@arthanson no worries! I added support for `PowerFeeds` as well.

arthanson (dismissed):

PR #712, opened 2025-11-13
Consider object_types as set in the response

### LINKED ISSUE #711: Synchronise list_as_set with ansible_module
### pynetbox version

v.7.5.0

### NetBox version

v.4.4.6

### Python version

3.10

### Steps to Reproduce

As documented in netbox-community/ansible_modules#1486 I would like to bringe the list_as_set variable in line with the ansible module.

Currently updating the netbox permissions via ansible results in reported changed cause the field object_types is compared as ordered list, not as set.

### Expected Behavior

as object_types doesn't need be ordered, it should be considered a set

### Observed Behavior

object_types are a generic list. Maybe other fields must be check as well.

dhoffend:
Fixes: #711

As documented in netbox-community/ansible_modules#1486 I would like to bringe the list_as_set variable in line with the ansible module.

Currently updating the netbox permissions via ansible results in reported changed cause the field object_types is compared as ordered list, not as set.

### DIFF
--- a/pynetbox/core/response.py
+++ b/pynetbox/core/response.py
@@ -23,7 +23,7 @@ from pynetbox.core.query import Request
 from pynetbox.core.util import Hashabledict
 
 # List of fields that are lists but should be treated as sets.
-LIST_AS_SET = ("tags", "tagged_vlans")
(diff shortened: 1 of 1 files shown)


arthanson (approved):
thanks @dhoffend !

Issue #505, opened 2022-10-17
Render Elevation as SVG


darcuripan:
I've managed to get the rack elevation as a list of devices
```
rack = netbox.dcim.racks.get(21)

device = rack.elevation.list()
```

and I know the REST API has the option to set render=json/svg, and I was wondering if there is an already existing way to do this in pynetbox or if it could be added in the future?

Thanks!

abhi1693:
What would be the use case to render the SVG using pynetbox?

cs-1:
We use pynetbox to create MS office documents as informational documents i. e. for contractors that need to work on racks etc. It would be great to be able to include the SVGs as visual guide in these documents.

NeverBehave:
If this is still a wanted feature please assign me on this. Thanks.

PR #720, opened 2025-12-09
feat(endpoint): add SVG support for rack elevation endpoint

NeverBehave:
### Fixes: #505

Add support for SVG elevation diagrams via render='svg' parameter on rack elevation endpoint.

### Changes

  - Add expect_json parameter to `Request.__init__()` (defaults to True)
  - Update `Request._make_call()` to set Accept header and parse responses based on expect_json
  - Create `ROMultiFormatDetailEndpoint` class for multiformat responses
  - Update Racks.elevation to use `ROMultiFormatDetailEndpoint`
  - Add documentation to `docs/advanced.md`

### DIFF
--- a/tests/unit/test_multiformat_endpoint.py
+++ b/tests/unit/test_multiformat_endpoint.py
@@ -138,7 +138,3 @@ class ROMultiFormatDetailEndpointTestCase(unittest.TestCase):
 
         self.assertIn("Unsupported render format", str(context.exception))
         self.assertIn("png", str(context.exception))
-
-
-if __name__ == "__main__":
-    unittest.main()
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -171,10 +171,3 @@ print(svg_diagram)  # '<svg xmlns="http://www.w3.org/2000/svg">...</svg>'
 with open('rack-elevation.svg', 'w') as f:
     f.write(svg_diagram)
 ```
-
-## Supported Formats
-
-The elevation endpoint supports:
-- `render='json'` - Explicitly request JSON (same as default)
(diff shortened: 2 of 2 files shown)


pynetbox/core/query.py
```diff
@@ -211,6 +212,16 @@ def __init__(
             In (e.g. /api/dcim/devices/?name='test') 'name': 'test'
             would be in the filters dict.
         * **key** (int, optional): Database id of the item being queried.
+        * **expect_json** (bool, optional): If True, expects JSON response
+            and sets appropriate Accept header. If False, expects raw content
+            (e.g., SVG, XML) and returns text. Defaults to True.
+
+        ## Note
```
arthanson: Is this supposed to be here, doesn't seem part of this PR?

tests/unit/test_multiformat_endpoint.py
```diff
@@ -0,0 +1,144 @@
...
+
+        # Unsupported render formats should raise ValueError
+        with self.assertRaises(ValueError) as context:
+            rack.elevation.list(render="png")
+
+        self.assertIn("Unsupported render format", str(context.exception))
+        self.assertIn("png", str(context.exception))
+
+
+if __name__ == "__main__":
```
arthanson: This would be for a standalone script not a test

pynetbox/core/query.py
NeverBehave: This is my note when exploring this part of the codebase. I am not entirely positive if I understand this correctly tho.
I tried to add an attribute and realize the `count` attribute was not declared but used, but it actually suppose to be that way.

arthanson (approved):

PR #736, opened 2026-01-12
#735 replace black with Ruff

### LINKED ISSUE #735: Replace black lining with Ruff
### Proposed Changes

To standardize project with NetBox, remove the black linter in GitHub actions and replace it with Ruff, also put in place a pre-commit hook to run the Ruff linter.

### Justification

Standardize the NetBox projects to make it easier to work between the applications, also Ruff is much faster and has a better baseline then Black and lints the file without reformatting them.

arthanson:
### Fixes: #735 

Replaced black with Ruff for linting to match NetBox and cleared up a Ruff error.

### DIFF
--- a/pynetbox/__init__.py
+++ b/pynetbox/__init__.py
@@ -11,7 +11,7 @@ __version__ = "7.6.0"
 # Lowercase alias for backward compatibility
 api = Api
 
-__all__ = [
+__all__ = (
     "Api",
     "AllocationError",
     "ContentError",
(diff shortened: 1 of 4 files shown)


pynetbox/__init__.py
```diff
@@ -10,3 +10,13 @@
 
 # Lowercase alias for backward compatibility
 api = Api
+
+__all__ = [
```
jeremystretch: Typically `__all__` is a tuple

jeremystretch (dismissed):
Approved w/some minor tweaks suggested

jeremystretch (approved):

PR #757, opened 2026-03-29
docs: Correct sidebar reference for IPAM

AlexCK-STFC:
Fixes: #755 

Correct 404 on sidebar reference to IPAM page by renaming to lower case, matching dcim.md


jeremystretch (commented):
Rather than altering the YAML config, could we instead rename the file itself to be consistent with `dcim.md`?

jeremystretch (approved):
Thanks @AlexCK-STFC!

PR #761, opened 2026-04-27
#756 use v2 token for create_token

arthanson:
### Fixes: #756

 NetBox 4.5 made v2 tokens the default. The provision endpoint now returns key (a short identifier) and token (the plaintext secret) separately; the full auth value must be constructed as nbt_<key>.<token>. create_token() was reading only resp["key"] and using it directly, which sent an invalid value in the Authorization header.

**Fix:** Detect version == 2 in the provision response and construct the full token as nbt_<key>.<token>. For v1 tokens, prefer the token field (NetBox 4.5+) and fall back to key (pre-4.5).

Also updated CI matrix for more modern netbox / python as this targets NetBox 4.5 specifically.  Required some changes to support new NetBox docker which uses granian instead of nginx.


arthanson:
@claude /review

pynetbox/core/api.py
pheus, line 251: Could we update the `create_token()` docstring example as part of this change?

The current example still shows only a legacy v1-style token value in `nb.token`. Since this method can now return/store a v2 auth value like `nbt_<key>.<token>`, a short note or updated example would help users understand why `nb.token` may differ from `token.key`.

tests/test_api.py
```diff
@@ -96,9 +96,36 @@ class ApiCreateTokenTestCase(unittest.TestCase):
         "requests.sessions.Session.post",
         return_value=Response(fixture="api/token_provision.json"),
```
pheus: Can we rename the old fixture for consistency?

```suggestion
        return_value=Response(fixture="api/token_provision_v1_legacy.json"),
```

pheus (changes requested):
Thanks for fixing this! This looks good to me overall. I only have a couple of minor suggestions.

Since this bug is specific to NetBox 4.5 token behavior, we may also want to add a NetBox 4.5 job to the CI matrix, so this behavior is covered by integration tests as well.

arthanson:
@pheus adding NetBox 4.5 actually caused a bunch of changes, NetBox Docker switched from nginx to granian which caused several changes and also there were lots of test changes needed for the v2 token and the front/rear port mappings.  Ready for re-review

pheus (approved):
Thanks for the follow-up changes! This looks good to me.

PR #771, opened 2026-05-12
#749 - fix: don't treat Record methods as constructors in _parse_values

### LINKED ISSUE #749: Cannot read an interface device config context data containing an "updates" key
### pynetbox version

v7.4.1

### NetBox version

v4.0.11

### Python version

3.12

### Steps to Reproduce

```python 
import pynetbox
import os

nb = pynebox.api('http://127.0.0.1:8000/', token='my-token')
device = nb.dcim.devices.create({
    "name": "test_device",
    "device_type": "my-device-type",
    "role": "my-role",
    "site": "my-site",
    "local_context_data": {"router_bgp": {"updates": {"wait_install": True}}}
})

iface = nb.dcim.interfaces.create({
    "device": device.id,
    "name": "Ethernet1/1",
    "type": "1000base-t",
})

print(device.local_context_data) # work
print(iface.device.local_context_data) # fail
```

### Expected Behavior

```python
{'router_bgp': {'updates': {'wait_install': True}}}
```

### Observed Behavior

```
  File "<stdin>", line 1, in <module>
  File "/home/vzlh8836/Landc/spark/.venv/lib/python3.12/site-packages/pynetbox/core/response.py", line 303, in __getattr__
    if self.full_details():
       ^^^^^^^^^^^^^^^^^^^
  File "/home/vzlh8836/Landc/spark/.venv/lib/python3.12/site-packages/pynetbox/core/response.py", line 459, in full_details
    self._parse_values(next(req.get()))
  File "/home/vzlh8836/Landc/spark/.venv/lib/python3.12/site-packages/pynetbox/core/response.py", line 408, in _parse_values
    v = self.default_ret(v, self.api, self.endpoint)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/vzlh8836/Landc/spark/.venv/lib/python3.12/site-packages/pynetbox/core/response.py", line 290, in __init__
    self._parse_values(values)
  File "/home/vzlh8836/Landc/spark/.venv/lib/python3.12/site-packages/pynetbox/core/response.py", line 408, in _parse_values
    v = self.default_ret(v, self.api, self.endpoint)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/vzlh8836/Landc/spark/.venv/lib/python3.12/site-packages/pynetbox/core/response.py", line 290, in __init__
    self._parse_values(values)
  File "/home/vzlh8836/Landc/spark/.venv/lib/python3.12/site-packages/pynetbox/core/response.py", line 406, in _parse_values
    v = lookup(v, self.api, self.endpoint)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: Record.updates() takes 1 positional argument but 3 were given
```
I cannot test with newer versions of netbox and pynetbox because of production's constraints, but the code seems to be same on the current master branch in `_parse_values` function.

arthanson:
### Fixes: #749 

- `Record._parse_values` looked up nested-dict field names with `getattr(self.__class__, k, None)` and invoked anything truthy as `lookup(v, api, endpoint)`. When a nested API payload contained a key matching a `Record` method name — e.g. `updates`, `save`, `delete`, `serialize` — the method was called as a constructor and raised `TypeError`.
- Narrowed the check to `isinstance(lookup, type)` so only class-typed field overrides take the constructor path; everything else falls through to `default_ret`. All in-tree model overrides in `pynetbox/models/*.py` are `Record` subclasses, so existing behavior is preserved.
- Added a regression test covering a nested `updates` key (the path that fails in the issue's repro).****

### DIFF
--- a/pynetbox/core/response.py
+++ b/pynetbox/core/response.py
@@ -482,7 +482,7 @@ class Record:
                     self._add_cache((k, copy.deepcopy(v)))
                     setattr(self, k, v)
                     continue
-                if lookup:
+                if isinstance(lookup, type):
                     v = lookup(v, self.api, self.endpoint)
                 else:
                     v = self.default_ret(v, self.api, self.endpoint)
--- a/pynetbox/core/response.py
+++ b/pynetbox/core/response.py
@@ -482,7 +482,7 @@ class Record:
                     self._add_cache((k, copy.deepcopy(v)))
                     setattr(self, k, v)
                     continue
-                if isinstance(lookup, type):
+                if isinstance(lookup, type) and issubclass(lookup, Record):
                     v = lookup(v, self.api, self.endpoint)
                 else:
                     v = self.default_ret(v, self.api, self.endpoint)
--- a/tests/unit/test_response.py
+++ b/tests/unit/test_response.py
@@ -52,6 +52,18 @@ class RecordTestCase(unittest.TestCase):
         with self.assertRaises(KeyError) as _:
             test_obj["nothing"]
 
+    def test_nested_dict_key_collides_with_record_method(self):
+        # Regression for issue #749: a nested dict key like "updates"
+        # collided with Record.updates and was invoked as a constructor.
+        test_values = {
+            "id": 1,
+            "config_context": {"router_bgp": {"updates": {"wait_install": True}}},
+        }
+        test_obj = Record(test_values, Mock(base_url="http://test"), None)
+        self.assertEqual(
+            test_obj.config_context.router_bgp.updates.wait_install, True
+        )
+
     def test_serialize_list_of_records(self):
         test_values = {
             "id": 123,


pynetbox/core/response.py
```diff
@@ -482,7 +482,7 @@ def list_parser(key_name, list_item):
                     self._add_cache((k, copy.deepcopy(v)))
                     setattr(self, k, v)
                     continue
-                if lookup:
+                if isinstance(lookup, type):
```
pheus:
```suggestion
                if isinstance(lookup, type) and issubclass(lookup, Record):
```

One small suggestion: could we make the predicate a bit more explicit and only instantiate `Record` subclasses?

pheus (dismissed):
Thanks! This looks good to me! I only have a non-blocking suggestion.

arthanson:
review changes requires re-review

pheus (approved):
Thanks! Looks good to me!

PR #775, opened 2026-05-18
#774 - Add plugin extension framework

### LINKED ISSUE #774: Add extension framework for NetBox plugins
### pynetbox version

v7.1.0

### NetBox version

v4.5.0

### Feature type

New functionality

### Proposed functionality

Introduce an opt-in extension framework that lets users (and plugin authors) register custom Record subclasses and content-type mappings for NetBox plugins, without modifying pynetbox itself.

An extension declares which plugin it targets (plugin_name), a namespace of Record subclasses to use for that plugin's endpoints, and an optional set of content-type mappings for polymorphic nested objects. Extensions are passed in explicitly at API construction time via a new extensions= kwarg on pynetbox.api(...).

When pynetbox resolves an endpoint under nb.plugins.<plugin_name>, it consults the registered extensions and uses the matching Record subclass — the same mechanism built-in apps already use, just opened up to plugins. Generic-FK parsing similarly honours extension-registered content types.

The change is fully additive: existing usage is unchanged, no auto-discovery, and the scope is limited to plugins under nb.plugins.* (built-in apps are not overridable).

### Use case

NetBox plugins frequently introduce endpoints, detail routes, and field shapes that don't fit the generic Record deserialization - JSON dict fields get mangled, custom detail endpoints (merge/sync/run/etc.) need bespoke methods, and plugin-specific content types aren't recognised by generic-FK parsing. Today, supporting any of this requires patching pynetbox itself and waiting on an upstream release, which doesn't scale to the growing plugin ecosystem.

This framework lets plugin authors ship companion packages with typed Record subclasses, detail endpoints, JSON-field markers, and content-type registrations; lets internal teams register extensions for in-house plugins from their own codebases; and gives pynetbox maintainers a cleaner answer to plugin-specific feature requests than absorbing every plugin's models into core.

### External dependencies

_No response_

arthanson:
### Fixes: #774 

Adds a plugin extension framework so NetBox plugins (or pynetbox itself) can register custom `Record` subclasses, JSON-field markers, and content-type mappings for `nb.plugins.<x>` endpoints — without modifying core pynetbox.

Foundation for the shipped extensions (e.g. #776 branching, #777 custom objects), and also usable by third parties who want typed records for their own plugins.

## Contract

An extension is any object exposing:

- **`plugin_name`** (str) — slug under `nb.plugins` (e.g. `"branching"` for `nb.plugins.branching`). Underscores map to dashes in the URL, matching the existing convention.
- **`models`** — a namespace (module, class, or instance) from which `Record` subclasses are resolved by title-cased endpoint name (e.g. `getattr(models, "Branches")` for `nb.plugins.branching.branches`).
- **`content_types`** (dict, optional) — maps NetBox content-type strings to `Record` subclasses so polymorphic nested objects deserialize correctly.

Subclassing `pynetbox.Extension` is optional; the base class exists to give authors an explicit contract.

## Usage

```python
import pynetbox
from pynetbox import Extension
from pynetbox.core.response import JsonField, Record


class Branches(Record):
    pass


class BranchingModels:
    Branches = Branches


class BranchingExtension(Extension):
    plugin_name = "branching"
    models = BranchingModels
    content_types = {"netbox_branching.branch": Branches}


nb = pynetbox.api(url, token="...", extensions=[BranchingExtension])
```

What it changes

- pynetbox/core/extension.py — new Extension base class.
- Api.__init__ — accepts extensions=[...]; builds per-instance _extensions and _content_type_mapper registries. Duplicate plugin_name or content-type keys raise ValueError; overriding a built-in CONTENT_TYPE_MAPPER entry is allowed.
- App._setmodel — looks up plugin models via the registry for plugins/<slug> apps.
- Record._parse_values and GenericListObject — use the per-instance content-type mapper instead of the global one, so extension records participate in GFK resolution.
- pynetbox/__init__.py — exports Extension.

### DIFF
--- a/pynetbox/core/extension.py
+++ b/pynetbox/core/extension.py
@@ -6,9 +6,15 @@ content-type mappings without modifying pynetbox itself.
 
 An extension is any object (typically a class) exposing:
 
-* ``plugin_name`` (str): the plugin's slug as it appears under ``nb.plugins``
-  (e.g. ``"branching"`` for ``nb.plugins.branching``). Underscores in the
-  attribute access map to dashes in the URL, matching the existing
+* ``plugin_name`` (str): the plugin's API URL slug as used under
+  ``/api/plugins/`` — *not* the plugin's Python package name or
+  ``PluginConfig.name``. For example, the ``netbox-acls`` plugin
+  serves ``/api/plugins/access-lists/`` so its ``plugin_name`` is
+  ``"access_lists"`` (or equivalently ``"access-lists"``).
+  Dashes are normalized to underscores on registration, so the
+  same plugin cannot be registered twice under both spellings.
+  The attribute access ``nb.plugins.<plugin_name>`` then converts
+  underscores back to dashes for the URL, matching the existing
   ``PluginsApp`` convention.
 * ``models``: a namespace (module, class, or any object) from which
   ``Record`` subclasses can be resolved by title-cased endpoint name —
--- a/pynetbox/core/extension.py
+++ b/pynetbox/core/extension.py
@@ -1,0 +1,62 @@
+"""
+Extension framework for pynetbox.
+
+Lets third-party NetBox plugins ship custom ``Record`` subclasses and
+content-type mappings without modifying pynetbox itself.
+
+An extension is any object (typically a class) exposing:
+
+* ``plugin_name`` (str): the plugin's slug as it appears under ``nb.plugins``
+  (e.g. ``"branching"`` for ``nb.plugins.branching``). Underscores in the
+  attribute access map to dashes in the URL, matching the existing
+  ``PluginsApp`` convention.
+* ``models``: a namespace (module, class, or any object) from which
+  ``Record`` subclasses can be resolved by title-cased endpoint name —
+  the same lookup used for built-in apps. For ``nb.plugins.<x>.branches``
+  pynetbox will look up ``models.Branches``.
+* ``content_types`` (dict, optional): maps NetBox content-type strings
+  (e.g. ``"netbox_branching.branch"``) to ``Record`` subclasses, so the
+  classes are used when resolving polymorphic nested objects.
+
+## Example
+
+```python
+from pynetbox import api, Extension, JsonField
+from pynetbox.core.response import Record
+from pynetbox.core.endpoint import DetailEndpoint
+
+
+class Branches(Record):
+    @property
+    def merge(self):
+        return DetailEndpoint(self, "merge")
+
+
+class BranchingModels:
+    Branches = Branches
+
+
+class BranchingExtension(Extension):
+    plugin_name = "branching"
+    models = BranchingModels
+
+
+nb = api("http://localhost:8000", token="...", extensions=[BranchingExtension])
+branch = nb.plugins.branching.branches.get(1)
+branch.merge.create(commit=True)
+```
+"""
+
+
+class Extension:
+    """Base class for pynetbox extensions.
+
+    Subclassing is optional — any object with the required attributes is
+    accepted. The base class exists so extension authors have an explicit
+    contract to target and so ``isinstance`` checks are available to
+    downstream code that wants them.
+    """
+
+    plugin_name: str = ""
+    models = None
+    content_types: dict = {}
--- a/pynetbox/core/response.py
+++ b/pynetbox/core/response.py
@@ -448,13 +448,17 @@ class Record:
         def generic_list_parser(key_name, list_item):
             from pynetbox.models.mapper import CONTENT_TYPE_MAPPER
 
+            content_type_mapper = getattr(
+                self.api, "_content_type_mapper", CONTENT_TYPE_MAPPER
+            )
+
             if (
                 isinstance(list_item, dict)
                 and "object_type" in list_item
                 and "object" in list_item
             ):
                 lookup = list_item["object_type"]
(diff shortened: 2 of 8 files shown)


arthanson:
@claude /review

pynetbox/core/extension.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+class BranchingExtension(Extension):
+    plugin_name = "branching"
+    models = BranchingModels
+
+
+nb = api("http://localhost:8000", token="...", extensions=[BranchingExtension])
+branch = nb.plugins.branching.branches.get(1)
+branch.merge.create(commit=True)
```
pheus, line 52: **Potential bug / test gap:** The extension example advertises usage like:

```python
branch.merge.create(commit=True)
```

but `DetailEndpoint.create()` currently only accepts a single `data=None` argument. As written, `branch.merge.create(commit=True)` will raise `TypeError` before making the request.

Could we either update the example to use the currently supported call style:

```python
branch.merge.create({"commit": True})
```

or extend `DetailEndpoint.create()` to accept keyword arguments in the same spirit as `Endpoint.create()`?

For example, something along these lines would preserve the existing dict-style call while supporting the documented branching usage:

```python
def create(self, data=None, **kwargs):
    if data is not None and kwargs:
        raise ValueError("Cannot pass both data and keyword arguments")
    payload = data if data is not None else kwargs

    req = Request(**self.request_kwargs).post(payload)
    if isinstance(req, dict):
        return self.return_obj(req, self.endpoint.api, self.endpoint)
    return req
```

I’d also add a small test that calls the detail endpoint with `commit=True`, since the current test only asserts that the detail endpoint URL is generated correctly.

pynetbox/core/extension.py
```diff
@@ -0,0 +1,62 @@
...
+        return DetailEndpoint(self, "merge")
+
+
+class BranchingModels:
+    Branches = Branches
+
+
+class BranchingExtension(Extension):
+    plugin_name = "branching"
+    models = BranchingModels
```
pheus, line 47: **Non-blocking API clarity:** I think the `plugin_name` contract should be documented very explicitly as the plugin API/base-url slug normalized for Python attribute access, not the Django plugin package name.

For example:

```python
class AccessListsExtension(Extension):
    # /api/plugins/access-lists/...
    plugin_name = "access_lists"
```

not:

```python
plugin_name = "netbox_acls"
```

Likewise:

```python
class BgpExtension(Extension):
    # /api/plugins/bgp/...
    plugin_name = "bgp"
```

Maybe the docstring could say something like:

> `plugin_name` is the plugin’s API URL slug as used under `/api/plugins/`, with hyphens written as underscores for pynetbox attribute access. It is not necessarily the plugin’s Python package name or `PluginConfig.name`.

It may also be worth normalizing/validating this on registration so `plugin_name = "custom-objects"` and `plugin_name = "custom_objects"` cannot silently behave as two different registrations.

pynetbox/core/response.py
```diff
@@ -448,13 +448,17 @@ def _parse_values(self, values):
         def generic_list_parser(key_name, list_item):
```
pheus, line 448: **Potential bug:** In `_parse_values()`, `generic_list_parser()` can return the original dict when the item’s `object_type` is not present in the content-type mapper. The caller then does:

```python
to_cache = [i.serialize() for i in v]
```

which assumes every parsed item is a `GenericListObject`.

With plugin extensions, users may reasonably register only some plugin content types, or receive API data from plugins they have not added an extension for yet. In that case, a generic relation list containing an unmapped plugin object would crash during parsing instead of preserving the raw API data.

Could we make the fallback tolerant here? For example:

```python
parsed = [generic_list_parser(k, i) for i in v]
v = parsed
to_cache = [
    i.serialize() if hasattr(i, "serialize") else copy.deepcopy(i)
    for i in parsed
]
```

A regression test with an unmapped `object_type`, for example `"some_plugin.some_model"`, would be useful. Expected behavior would be that mapped entries become `GenericListObject` instances and unmapped entries remain as raw dictionaries rather than causing an exception.

pheus (changes requested):
Thanks for working on this! The extension registry approach looks like the right direction to me.

I’d like to see two small correctness items addressed before merge:

1. Either make the branching-style `branch.merge.create(commit=True)` example work, or update the example/tests to use the currently supported `create({"commit": True})` style.
2. Make generic-list parsing tolerant of unmapped plugin content types so a partially registered plugin extension does not crash on `i.serialize()`.

The `plugin_name`/plugin slug naming could be handled as documentation/API polish, but I think it is worth clarifying in this PR because real NetBox plugins often have different Python package names and API base URLs.


**Follow-up / separate issue:** This extension framework should help with static Custom Objects endpoints like `custom-object-type-fields`, but dynamic Custom Objects endpoints may still run into pynetbox’s normal underscore-to-hyphen endpoint conversion.

NetBox Custom Objects can expose endpoints based on the custom object type name, including names with underscores such as:

```text
/api/plugins/custom-objects/dhcp_scope/
```

But pynetbox currently converts:

```python
nb.plugins.custom_objects.dhcp_scope
```

to:

```text
/api/plugins/custom-objects/dhcp-scope/
```

I don’t think this necessarily needs to be solved in this PR, but it would be good to document as out of scope or cover in the companion Custom Objects extension PR.

pheus (approved):
Thanks! Nice work, @arthanson !

PR #783, opened 2026-06-04
#688: Skip network fetch for dunder attributes in Record.__getattr__

### LINKED ISSUE #688: Unexpected re-hydration on isinstance-check
### pynetbox version

v.7.4.1

### NetBox version

v4.2.6

### Python version

3.11

### Steps to Reproduce

When doing isinstance-checks on a Record, there's a chance of re-hydration happening, with two unwanted effects:
1) An unnecessary network call is made
2) Any updates done on the Record is reset

For instance, the internal instance check of BaseModel of [Pydantic](https://docs.pydantic.dev/latest/), will perform a `hasattr(instance, '__pydantic_decorators__')` on the Record, triggering `Record.__getattr__` and thus a re-hydration.

Example:
```python
import pynetbox
from pydantic import BaseModel

nb = pynetbox.api(...)

device = nb.dcim.devices.get(1) # Any object works

device.name = "New Name"

isinstance(device, BaseModel) # Will trigger network call and re-hydration/resetting of values

assert device.name != "New Name" # Local name update will have been reset
```

### Expected Behavior

Local updates on the Record should persist after \_\_getattr\_\_ is called by overloaded instance checks. Furthermore, not doing a network call in this case would be preferable.

Specifically, I'm suggesting the following change to `Record.__getattr__`.

```python
def __getattr__(self, k):
    if self.url:
        if self.has_details is False and k != "keys" and not k.startswith("__"): # <---- Fix 1: No need for network calls for dunder attrs
            if self.full_details():
                ret = getattr(self, k, None)
                if ret or hasattr(self, k):
                    return ret

    raise AttributeError('object has no attribute "{}"'.format(k))
```

along with the following change to `Record.full_details`

```python
def full_details(self):
    if self.url:
        req = Request(
            base=self.url,
            token=self.api.token,
            http_session=self.api.http_session,
        )
        curr_updates = self.updates()        # <---- Fix 2: Save current updates
        self._parse_values(next(req.get()))
        for k, v in curr_updates.items():    # <---- Fix 2: Restore
            setattr(self, k, v)              # <---- current updates
        self.has_details = True
        return True
    return False
```

### Observed Behavior

See code comments in "Steps to Reproduce"

arthanson:
### Fixes: #688 

short-circuit dunder attributes at the top of __getattr__ — they're never NetBox fields, so they raise AttributeError immediately without any fetch. This also protects against copy.deepcopy, pickling, and other libraries that probe for dunders.

### DIFF
--- a/pynetbox/core/response.py
+++ b/pynetbox/core/response.py
@@ -334,7 +334,16 @@ class Record:
 
         In order to prevent non-explicit behavior,`k='keys'` is
         excluded because casting to dict() calls this attr.
+
+        Dunder attributes are excluded as well: they are never NetBox
+        fields, and probing for them (e.g. pydantic's isinstance check
+        calling `hasattr(obj, '__pydantic_decorators__')`, or copy and
+        pickle machinery) must not trigger a network fetch that would
+        also clobber local modifications.
         """
+        if k.startswith("__") and k.endswith("__"):
+            raise AttributeError('object has no attribute "{}"'.format(k))
+
         if self.url:
             if self.has_details is False and k != "keys":
                 if self.full_details():
--- a/tests/unit/test_response.py
+++ b/tests/unit/test_response.py
@@ -37,6 +37,24 @@ class RecordTestCase(unittest.TestCase):
         with self.assertRaises(AttributeError) as _:
             test_obj.nothing
 
+    def test_dunder_attribute_does_not_trigger_full_details(self):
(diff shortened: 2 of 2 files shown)


arthanson:

arthanson:
Claude comment fixed.

pheus (approved):
Thanks!

PR #784, opened 2026-06-04
#715 - allow endpoints with underscores

### LINKED ISSUE #715: Underscores in slugs are converted to hyphens
### pynetbox version

v7.5.0

### NetBox version

v4.4.5

### Python version

3.11

### Steps to Reproduce

I am using netbox 4.4.x and the custom objects plugins and I am running in an issue exactly like #653 

I tend to name my objects with underscores (as in object_for_something), probably because I do too much python, and it prevents pynetbox from working there as it converts underscores to hyphens.
This is easily visible in the request part of a pynetbox object.

I am not sure if pynetbox should be flexible, or if netbox should enforce - instead of _ in slugs ?

It's pretty easy to reproduce, create a custom object and name the slug with an underscore, then try to call it via pynetbox.

Happy to provide more details if needed.

### Expected Behavior

If I call my object that has underscores in the slug, I expect the lib to reflect the underscores in the naming : 
'https://netbox/api/plugins/custom-objects/my_first_custom_object/'

### Observed Behavior

The url for the request is converted like so : 
'https://netbox/api/plugins/custom-objects/my-first-custom-object/'

arthanson:
### Fixes: #715

Adds a new App.endpoint(name) method that constructs an Endpoint using name as the literal URL slug, skipping the underscore-to-dash conversion:

```
# Attribute access — underscores become dashes (unchanged):
nb.plugins.custom_objects.my_first_custom_object        # → .../custom-objects/my-first-custom-object/

# New escape hatch — slug used verbatim:
nb.plugins.custom_objects.endpoint("my_first_custom_object").all()   # → .../custom-objects/my_first_custom_object/
```

- Opt-in and per-call — no global state, thread-safe, and zero change to existing attribute-access behavior.
- Generic — works on any App/plugin, not just custom objects, so it doesn't special-case any prefix.
- Mapping preserved — endpoint() passes the same model as attribute access, so model/return-object resolution is unchanged (e.g. you still get typed CustomObject records with the extension registered).

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -53,6 +53,7 @@ The `App` class represents a NetBox application (such as `dcim`, `ipam`, or `cir
     options:
         members:
             - config
+            - endpoint
         show_source: true
         show_root_heading: true
         heading_level: 3
--- a/docs/custom-objects.md
+++ b/docs/custom-objects.md
@@ -54,6 +54,20 @@ new = nb.plugins.custom_objects.cidr_list.create(name="ours", prefix=42)
 
 `CustomObject.custom_object_type` resolves to a `CustomObjectTypes` record, so `obj.custom_object_type.slug` works without an extra fetch when the per-type serializer includes it.
 
+### Slugs that contain underscores
+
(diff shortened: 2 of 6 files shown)


arthanson:

arthanson:
Claude review - 1 not really an issue, 2 is very minor - if it ever comes up (which I doubt) can add something like it later.

bctiemann (approved):

PR #785, opened 2026-06-04
#748 - Stop re-sending PATCH for unchanged partial custom_fields

### LINKED ISSUE #748: update always updates if custom_fields are set
### pynetbox version

v7.6.1

### NetBox version

v4.5.2

### Python version

3.12

### Steps to Reproduce

Create device with some custom_field
get device()
call update() with custom_fields set the same as in create()
function return True in diff there is "custom_fields" set

### Expected Behavior

update() should return False if custom_fields did not change

### Observed Behavior

calling update() with custom_fields set always call api even if custom_fields was not changed

arthanson:
### Fixes: #748 

update()/save() issued a redundant PATCH (returning True) whenever a subset of an object's custom fields was passed with unchanged values - only visible when the object had more than one custom field defined. 

NetBox uses merge semantics for custom-field PATCHes (omitted keys are left untouched), but _diff() compared the full custom_fields dicts, so a partial assignment dropped the other keys and was wrongly flagged as changed. 

This change restricts the custom_fields comparison in _diff() to keys present in the new value; real value changes, new keys, and explicit None clears are still detected. Adds regression tests for the no-op and real-change cases.

### DIFF
--- a/pynetbox/core/response.py
+++ b/pynetbox/core/response.py
@@ -694,7 +694,9 @@ class Record:
         # field omitted from an update unchanged. Restrict the comparison to the
         # keys present in the current value so that assigning a subset of custom
         # fields isn't seen as removing the others (issue #748). To clear a
-        # custom field the caller still sets it explicitly to None.
+        # custom field the caller still sets it explicitly to None. Assigning an
+        # empty dict is therefore a no-op (it touches no keys), matching NetBox's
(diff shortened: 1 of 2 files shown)


arthanson:

arthanson:
Added test cases per Claude's review. For 2 just update doc-comment.

pheus (approved):
Looks good to me! Thanks!

PR #786, opened 2026-06-04
#633 - Support custom thread pool executor and configurable max_workers

### LINKED ISSUE #633: Support for custom threadpool
### pynetbox version

v7.3.4

### NetBox version

v3.6.0

### Feature type

Change to existing functionality

### Proposed functionality

The ability to inject a custom threadpool implementation into the Api object so users of have a more control over the behaviour.

The threadpool can be an extra, optional, argument to the Api constructor and usage would require a single line change

```
    def concurrent_get(self, ret, page_size, page_offsets):
        futures_to_results = []
           with self.threadpool(max_workers=4) as pool:
                ...
```

(from https://github.com/netbox-community/pynetbox/blob/d62cb871ca64cd38ae496292d0ecb6d84e257d70/pynetbox/core/query.py#L261C1-L263C59)

### Use case

I have a custom API that uses netbox as a backend through pynetbox. This API has tracing support, which generally works well with the automatic `requests` (the python lib) instrumentation from open telemetry.
However the tracing breaks when enabling the threading feature, as the trace id needs to be injected into the thread (contextvar I believe).

The threadpool can still be overwritten but it highly depends on internal usage and naming:

```
    import pynetbox.core.query

    # replace threadpool with a tracing enabled one
    import myapp.tracing.netbox_futures as nf
```

and `netbox_futures.py`
```
import concurrent.futures
from concurrent.futures import *  # noqa
from typing import Callable

from opentelemetry import context as otel_context
from opentelemetry.trace import get_tracer

class ThreadPoolExecutor(concurrent.futures.ThreadPoolExecutor):  # type: ignore
  ...
```

### External dependencies

None

arthanson:
### Fixes: #633

Adds two optional, backward-compatible parameters to `Api()` that let callers control how threaded `.all()` / `.filter()` queries fetch pages:

- **`thread_pool_executor`** - a `ThreadPoolExecutor` class or factory used to build the pool. Defaults to the stdlib `concurrent.futures.ThreadPoolExecutor`.
- **`max_workers`** - exposes the previously hardcoded worker count (`4`).

## Motivation

The stdlib `ThreadPoolExecutor` does not propagate the calling thread's `contextvars` / thread-locals into worker threads, so anything relying on that breaks silently when `threading=True`: OpenTelemetry/distributed traces detach, structured-logging correlation IDs vanish, and APM instrumentation misses the threaded work. Injecting a context-aware executor fixes this. Exposing `max_workers` also lets users tune concurrency for their NetBox instance.

## Changes

- `core/api.py` - accept and store `thread_pool_executor` and `max_workers`.
- `core/endpoint.py` - pass both into the `Request` built by `.all()` / `.filter()`.
- `core/query.py` - `concurrent_get` now uses `self.thread_pool_executor(max_workers=self.max_workers)` instead of a hardcoded `ThreadPoolExecutor(max_workers=4)`.

## Compatibility

Fully backward compatible - defaults reproduce today's exact behavior. Only the `threading=True` page-fetching path is affected.

## Tests & docs

- Unit tests for default resolution, custom executor invocation with the configured `max_workers`, and propagation from `Api` into `Request`.
- `docs/advanced.md`: new sections on tuning worker count and supplying a custom executor (with an OpenTelemetry context-propagation example).

### DIFF
--- a/docs/advanced.md
+++ b/docs/advanced.md
@@ -53,6 +53,61 @@ devices = list(nb_threaded.dcim.devices.all())
 print(f"With threading: {time.time() - start:.2f}s")
 ```
 
+### Tuning the Worker Count
+
+By default, threaded queries use up to 4 worker threads. Override this with `max_workers`:
+
+```python
+nb = pynetbox.api(
+    'http://localhost:8000',
+    token='your-token',
+    threading=True,
+    max_workers=8,
+)
+```
+
+### Custom Thread Pool Executor
+
+By default pynetbox builds its pool with `concurrent.futures.ThreadPoolExecutor`. You can
+inject your own executor with `thread_pool_executor`. It must be a callable matching the
+`ThreadPoolExecutor(max_workers=...)` signature and support the context-manager protocol —
+a `ThreadPoolExecutor` subclass is the simplest option.
+
+This is useful when worker threads need thread-local state that the standard executor does
+not propagate, such as an OpenTelemetry trace context, a request-scoped logging correlation
+ID, or a thread-bound database/session handle. Without propagation, work performed in the
+pool's threads is detached from the context of the calling thread.
+
+```python
+import concurrent.futures
+from opentelemetry import context as otel_context
+
+class ContextPropagatingExecutor(concurrent.futures.ThreadPoolExecutor):
+    """Carries the caller's OpenTelemetry context into each worker thread."""
+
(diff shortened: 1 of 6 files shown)


arthanson:

arthanson:
Claude review issues fixed.

bctiemann (approved):

PR #787, opened 2026-06-05
#764 - Support cursor-based pagination from NetBox 4.6

### LINKED ISSUE #764: Support cursor-based pagination from NetBox 4.6
### pynetbox version

v7.5.0

### NetBox version

v4.6.0-beta2

### Feature type

Change to existing functionality

### Proposed functionality

From NetBox 4.6.0-beta2:

https://github.com/netbox-community/netbox/issues/21363

> Implement cursor-based pagination for REST API, similar to what we did for the GraphQL API under https://github.com/netbox-community/netbox/issues/21110 in v4.5.2. This entails introducing a new `start` query parameter as an alternative to the `offset` parameter. This would be used in conjunction with the `limit` parameter` to paginate results [...]

Implement this in Pynetbox when `threading=True`.

### Use case

Performance improvement, as described in the NetBox issue above:

> The current pagination approach relies on a relative offset. This struggles with very large result sets, because the offset must be calculated by scanning the entire table up to the offset position. Employing an object's unique primary key as a cursor avoids this, as we can leverage simple numeric comparison to filter for all PKs greater than or equal to a given value. Combined with the existing `limit` capability, this affords very efficient pagination at the cost of a fixed (and not very meaningful) ordering.

### External dependencies

NetBox 4.6+

arthanson:
### Fixes: #764 

Adds opt-in cursor-based pagination for .all() and .filter(). NetBox 4.6 introduced cursor pagination as an alternative to offset pagination: instead of skipping a growing number of rows with offset, the server pages forward using the primary key as a cursor, which performs significantly better on very large result sets.

## Usage
```
nb = pynetbox.api(url, token=token, pagination="cursor")
devices = nb.dcim.devices.all()  # pages via the server's `start` cursor, following `next` links
```

pagination defaults to "offset", so existing behavior is unchanged.

Below is a script that shows the new pagination, will need to update the token.

 <details>
  <summary>Manual verification script</summary>

  ```python
#!/usr/bin/env python3
"""Manual test for cursor-based pagination (NetBox 4.6+, PR #787).

Exercises the cursor pagination path of ``.all()`` / ``.filter()`` against a
live NetBox instance and verifies the review-comment fixes:

  1. ``pagination="cursor"`` walks every page by following the server's
     ``next`` links and returns the same objects as offset pagination.
  2. ``len()`` on a cursor RecordSet refetches the count NetBox omits (null)
     in cursor mode.
  3. ``threading=True`` + ``pagination="cursor"`` emits a UserWarning because
     cursor pages are fetched sequentially (the threading config is a no-op).
  4. A transport-level failure in the version probe falls back to offset
     instead of propagating (api.py exception handling).

Requires a NetBox 4.6+ instance whose ``dcim.devices`` endpoint has more than
one page of results (the script forces a small page size to guarantee this).

Usage: python manual_tests/test_cursor_pagination.py
"""

import os
import sys
import warnings
from unittest.mock import patch

import requests

sys.path.insert(0, os.path.dirname(os.path.dirname(os.path.abspath(__file__))))

import pynetbox
from packaging import version

NETBOX_URL = os.environ.get("NETBOX_URL", "http://127.0.0.1:8000/")
NETBOX_TOKEN = "nbt_cUOdzNJvWx3q.v1YQC2Q1smP0PyN4soKGucD3BrbYZcu1qyztXu2c"

# Small page size so the device list spans multiple cursor pages and the
# next-link following is actually exercised.
PAGE_SIZE = 2


def test_version_and_effective_pagination():
    print("1. server version + effective pagination resolution")
    api = pynetbox.api(NETBOX_URL, token=NETBOX_TOKEN, pagination="cursor")
    server_version = api.version
    print(f"   NetBox {server_version} at {NETBOX_URL}")
    if version.parse(server_version) < version.parse("4.6"):
        raise SystemExit(
            f"   NetBox {server_version} does not support cursor pagination "
            "(need 4.6+); aborting."
        )
    effective = api._effective_pagination()
    assert effective == "cursor", f"expected 'cursor', got {effective!r}"
    print("   _effective_pagination() == 'cursor'")
    return api


def test_cursor_matches_offset(cursor_api):
    print("2. cursor pagination returns the same devices as offset")
    offset_api = pynetbox.api(NETBOX_URL, token=NETBOX_TOKEN, pagination="offset")

    offset_devices = list(offset_api.dcim.devices.all())
    cursor_devices = list(cursor_api.dcim.devices.filter(limit=PAGE_SIZE))

    print(f"   offset returned {len(offset_devices)} device(s)")
    print(f"   cursor returned {len(cursor_devices)} device(s) (page size {PAGE_SIZE})")

    assert len(cursor_devices) > PAGE_SIZE, (
        f"need more than one page of devices to test pagination; got "
        f"{len(cursor_devices)} with page size {PAGE_SIZE}"
    )

    offset_ids = sorted(d.id for d in offset_devices)
    cursor_ids = sorted(d.id for d in cursor_devices)
    assert offset_ids == cursor_ids, (
        "cursor and offset returned different device id sets:\n"
        f"   only in offset: {set(offset_ids) - set(cursor_ids)}\n"
        f"   only in cursor: {set(cursor_ids) - set(offset_ids)}"
    )
    print(f"   both strategies returned the same {len(cursor_ids)} device id(s)")


def test_cursor_len_refetches_count(cursor_api):
    print("3. len() on a cursor RecordSet refetches the omitted count")
    record_set = cursor_api.dcim.devices.filter(limit=PAGE_SIZE)
    count = len(record_set)
    print(f"   len(filter(limit={PAGE_SIZE})) == {count}")
    assert count > PAGE_SIZE, f"expected count > {PAGE_SIZE}, got {count}"


def test_threading_cursor_warns():
    print("4. threading=True + cursor emits a no-op warning")
    api = pynetbox.api(
        NETBOX_URL, token=NETBOX_TOKEN, pagination="cursor", threading=True
    )
    with warnings.catch_warnings(record=True) as caught:
        warnings.simplefilter("always")
        effective = api._effective_pagination()
    assert effective == "cursor", f"expected 'cursor', got {effective!r}"
    messages = [str(w.message) for w in caught if issubclass(w.category, UserWarning)]
    assert any("threading" in m for m in messages), (
        f"expected a threading no-op UserWarning, got: {messages}"
    )
    print(f"   warned: {messages[0]}")


def test_version_probe_network_error_falls_back():
    print("5. transport error in version probe falls back to offset")
    api = pynetbox.api(NETBOX_URL, token=NETBOX_TOKEN, pagination="cursor")
    with patch(
        "requests.sessions.Session.get",
        side_effect=requests.exceptions.ConnectionError("boom"),
    ):
        effective = api._effective_pagination()
    assert effective == "offset", f"expected 'offset', got {effective!r}"
    print("   ConnectionError during probe -> _effective_pagination() == 'offset'")


def main():
    api = test_version_and_effective_pagination()
    test_cursor_matches_offset(api)
    test_cursor_len_refetches_count(api)
    test_threading_cursor_warns()
    test_version_probe_network_error_falls_back()
    print("PASS")


if __name__ == "__main__":
    sys.exit(main())
  ```

  </details>

### DIFF
--- a/pynetbox/core/api.py
+++ b/pynetbox/core/api.py
@@ -255,10 +255,16 @@ class Api:
                 # be parallelised; the cursor path ignores self.threading.
                 # Warn so the no-op threading configuration is not a silent
                 # performance surprise.
+                # stacklevel=5 attributes the warning to the caller's list
+                # request rather than to pynetbox internals. The version probe
+                # is resolved lazily on the first page fetch, so the fixed
+                # frame chain at this point is:
+                #   _effective_pagination -> Request._resolve_pagination
+                #   -> Request.get -> RecordSet.__next__/__len__ -> caller.
                 warnings.warn(
                     "threading=True has no effect with cursor pagination; "
                     "cursor pages are fetched sequentially.",
-                    stacklevel=2,
+                    stacklevel=5,
                 )
         return "cursor" if self._cursor_supported else "offset"
 
--- a/pynetbox/core/api.py
+++ b/pynetbox/core/api.py
@@ -15,6 +15,7 @@ limitations under the License.
 """
 
 import contextlib
+import warnings
 
 import requests
 from packaging import version
@@ -220,8 +221,24 @@ class Api:
                 self._cursor_supported = version.parse(self.version) >= version.parse(
                     "4.6"
                 )
-            except (RequestError, InvalidVersion):
+            except (RequestError, InvalidVersion, requests.exceptions.RequestException):
+                # RequestError covers a non-ok HTTP response from the version
+                # probe; requests.exceptions.RequestException covers transport
+                # failures (ConnectionError, Timeout, ...) raised before a
+                # response exists. In every case fall back to offset, as the
+                # docstring promises, and let the real list request surface any
+                # underlying connectivity error.
                 self._cursor_supported = False
+            if self._cursor_supported and self.threading:
+                # Cursor pagination follows next links sequentially and cannot
+                # be parallelised; the cursor path ignores self.threading.
+                # Warn so the no-op threading configuration is not a silent
+                # performance surprise.
+                warnings.warn(
+                    "threading=True has no effect with cursor pagination; "
+                    "cursor pages are fetched sequentially.",
+                    stacklevel=2,
+                )
         return "cursor" if self._cursor_supported else "offset"
 
     def openapi(self):
--- a/pynetbox/core/api.py
+++ b/pynetbox/core/api.py
@@ -17,9 +17,11 @@ limitations under the License.
 import contextlib
 
 import requests
+from packaging import version
+from packaging.version import InvalidVersion
 
 from pynetbox.core.app import App, PluginsApp
-from pynetbox.core.query import Request, TOKEN_PREFIX
+from pynetbox.core.query import Request, RequestError, TOKEN_PREFIX
 from pynetbox.core.response import Record
 from pynetbox.models.mapper import CONTENT_TYPE_MAPPER
 
@@ -82,6 +84,7 @@ class Api:
         threading=False,
         strict_filters=False,
         extensions=None,
+        pagination="offset",
     ):
         """Initialize the API client.
 
@@ -91,13 +94,20 @@ class Api:
             threading (bool, optional): Set to True to use threading in `.all()` and `.filter()` requests, defaults to False.
             strict_filters (bool, optional): Set to True to check GET call filters against OpenAPI specifications (intentionally not done in NetBox API), defaults to False.
             extensions (list, optional): A list of `Extension` classes or instances that register custom `Record` subclasses and content-type mappings for NetBox plugins. See `pynetbox.core.extension`.
+            pagination (str, optional): Pagination strategy for `.all()` and `.filter()`, either `"offset"` (default) or `"cursor"`. Cursor pagination (NetBox 4.6+) offers better performance on very large result sets but omits the total count and cannot be combined with threading or `ordering`. On NetBox versions older than 4.6 it transparently falls back to offset pagination.
         """
+        if pagination not in ("offset", "cursor"):
+            raise ValueError(
+                "pagination must be 'offset' or 'cursor', got {!r}".format(pagination)
+            )
         base_url = "{}/api".format(url if url[-1] != "/" else url[:-1])
         self.token = token
         self.base_url = base_url
         self.http_session = requests.Session()
         self.threading = threading
         self.strict_filters = strict_filters
+        self.pagination = pagination
+        self._cursor_supported = None
 
         self._register_extensions(extensions or [])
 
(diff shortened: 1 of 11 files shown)


arthanson:

arthanson:
claude review issues fixed.

jeremystretch:

tests/integration/test_pagination.py
```diff
@@ -0,0 +1,56 @@
...
+        self, cursor_api, nb_version, prefixes
+    ):
+        """Cursor pagination walks every page and yields all objects.
+
+        Gated to NetBox 4.6+, where cursor pagination is available. On older
+        versions the client transparently falls back to offset pagination, so
+        this end-to-end check of the ``start`` semantics and ``next`` link
+        following is only meaningful from 4.6 onward.
+        """
+        if nb_version < version.parse("4.6"):
```
jeremystretch, line 60: CI currently only runs against NetBox v4.3-4.5. Probably worth adding v4.6 to ensure these tests run.

pynetbox/core/endpoint.py
```diff
@@ -190,6 +190,7 @@ def all(self, limit=0, offset=None):
             max_workers=self.api.max_workers,
             limit=limit,
             offset=offset,
+            pagination=self.api._effective_pagination(),
```
jeremystretch: Probably a minor concern, but calling `_effective_pagination()` on init triggers the `version` lookup, which causes an initial API request, even if pagination ultimately isn't needed to satisfy the request. Might be overkill, but what do you think about using a lazy object to defer the version lookup until it's actually needed?

pynetbox/core/api.py
```diff
@@ -215,6 +226,42 @@ def version(self):
...
+                self._cursor_supported = False
+            if self._cursor_supported and self.threading:
+                # Cursor pagination follows next links sequentially and cannot
+                # be parallelised; the cursor path ignores self.threading.
+                # Warn so the no-op threading configuration is not a silent
+                # performance surprise.
+                warnings.warn(
+                    "threading=True has no effect with cursor pagination; "
+                    "cursor pages are fetched sequentially.",
+                    stacklevel=2,
```
jeremystretch: Per [Claude's feedback](https://github.com/netbox-community/pynetbox/pull/787#issuecomment-4693435200)
```suggestion
                    stacklevel=3,
```

jeremystretch (changes requested):

tests/integration/test_pagination.py
```diff
@@ -0,0 +1,56 @@
...
+    created = [
+        api.ipam.prefixes.create(prefix="198.51.100.{}/32".format(i))
+        for i in range(5)
+    ]
+    yield created
+    for prefix in created:
+        prefix.delete()
+
+
+class TestCursorPagination:
```
jeremystretch, line 49: As written, these tests could potentially pass even if cursor pagination isn't actually being used. We should inspect the request and assert that the `start` parameter is being included for more robust coverage.

pynetbox/core/api.py
```diff
@@ -215,6 +226,48 @@ def version(self):
...
+                )
+            except (RequestError, InvalidVersion, requests.exceptions.RequestException):
+                # RequestError covers a non-ok HTTP response from the version
+                # probe; requests.exceptions.RequestException covers transport
+                # failures (ConnectionError, Timeout, ...) raised before a
+                # response exists. In every case fall back to offset, as the
+                # docstring promises, and let the real list request surface any
+                # underlying connectivity error.
+                self._cursor_supported = False
+            if self._cursor_supported and self.threading:
```
jeremystretch, line 253: Can we also raise a warning if `ordering` has been specified while cursor pagination is in use?

jeremystretch (dismissed):
Approved with two minor suggestions.

arthanson:
Added the two changes, so needs re-review.

jeremystretch (approved):

PR #794, opened 2026-08-17
Closes #170: Prefix internal Record attrs to avoid collision with custom fields

### LINKED ISSUE #170: Record attributes can conflict with NetBox model fields.
I discovered this problem using pynetbox to retrieve records for a custom model that coincidentally has an `endpoint` field.

Admittedly, a `Record` only has a handful of private attributes, and it is unlikely that any would conflict with core NetBox model fields, but it would be safer and definitely more future proof if they were prefixed with `_`.

salocin24:
### Fixes: #170 

Renames `Record`'s internal bookkeeping attributes (`api`, `endpoint`,
`has_details`, `default_ret`) to underscore-prefixed versions
(`_api`, `_endpoint`, `_has_details`, `_default_ret`) so they can no
longer be silently overwritten by a NetBox custom field that happens
to share the same name (e.g. a custom field literally named
`endpoint`).

To avoid a breaking change for any code currently accessing these
attributes directly, the original names are preserved as read-only
`@property` aliases on `Record` (e.g. `record.endpoint` still works,
now returning `self._endpoint`).

`url` was deliberately left unprefixed unlike the others, it's a NetBox API field, not internal bookkeeping, and is meant to be overwritten by API response data.

## Testing

- Added a regression test
  (`test_field_named_endpoint_does_not_corrupt_internal_endpoint`)
  that reproduces the exact scenario from #170: a field named
  `endpoint` now raises `AttributeError` on the read-only property
  instead of silently corrupting the Record's internal endpoint
  reference.

### DIFF
--- a/pynetbox/core/response.py
+++ b/pynetbox/core/response.py
@@ -332,28 +332,44 @@ class Record:
     # These are object bookkeeping attributes, not NetBox API fields.
     _INTERNAL_ATTRS = frozenset(
         [
-            "api",  # API client instance
-            "endpoint",  # Endpoint object reference
+            "_api",  # API client instance
+            "_endpoint",  # Endpoint object reference
             "url",  # Object URL (read-only field provided by API)
-            "has_details",  # Flag for lazy-loading full details
-            "default_ret",  # Default Record class for nested objects
+            "_has_details",  # Flag for lazy-loading full details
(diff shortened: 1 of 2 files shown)


pheus:
The lint failures are unrelated to this PR and will be addressed as part of the repository-wide Ruff cleanup tracked in #798.

